PR #25, opened 2017-10-25
Management locks API in fluent

martinsawicki:


martinsawicki:
@anuchandy if you could look at especially the collection methods, that'd be appreciated.

PR #29, opened 2017-10-31
Zone support for application gateway

anuchandy:
* Gateway zone support

* Ensuring gateway zone uniqueness & update java doc


anuchandy:
@jianghaolu this feature was in the private and we decided to move it to public. Could you please take a look and ensure this is not messing our clean history?

anuchandy:
@martinsawicki please review..

PR #32, opened 2017-11-06
Sample for Express Route and enabling gateway creation for Express Route

lenala:


azure-mgmt-network/src/main/java/com/microsoft/azure/management/network/VirtualNetworkGatewayConnection.java
```diff
@@ -144,7 +145,7 @@
 
             WithSecondVirtualNetworkGateway withVNetToVNet();
 
-            WithExpressRoute withExpressRoute();
+            WithExpressRouteCircuit withExpressRoute();
```
martinsawicki: missing javadocs for all these withers and interfaces. (I wonder why this passed checkstyle)

PR #40, opened 2017-11-08
Sample: Making use of RX in complex virtual machine creation scenarios

martinsawicki:
(based on a similar unit yes we've had for a while)


azure-samples/src/main/java/com/microsoft/azure/management/compute/samples/CreateVirtualMachinesAsyncTrackingRelatedResources.java
```diff
@@ -0,0 +1,321 @@
...
+                            } else {
+                                // Since this is not a VM, add this resource to the potential cleanup list
+                                createdResourceIds.put(resource.key(), resource.id());
+                            }
+                        }
+                        return createdResource;
+                    }
+               })
+               .onErrorReturn(new Func1<Throwable, Indexable>() {
+                   // Aggregate all the errors
```
anuchandy: There shouldn't be multiple errors. The upstream observable should emit error only once, 

(1). if there are multiple internal errors during the parallel execution then upstream emits rx.CompositeException which aggregate all errors 
(2). if there is only one exception during parallel execution then it will be emitted as it is.

So in nutshell we can expect only one error if there is any.

azure-samples/src/main/java/com/microsoft/azure/management/compute/samples/CreateVirtualMachinesAsyncTrackingRelatedResources.java
```diff
@@ -0,0 +1,321 @@
...
+                   @Override
+                   public Indexable call(Throwable throwable) {
+                       errors.add(throwable);
+                       return null;
+                   }
```
anuchandy: Since it is guaranteed to emit only one error (rx.CompositeException or single Throwable) if there is any, we can use `.doOnError()` instead of `onErrorReturn ` and store error in a single error variable as oppose to in a list.

azure-samples/src/main/java/com/microsoft/azure/management/compute/samples/CreateVirtualMachinesAsyncTrackingRelatedResources.java
```diff
@@ -0,0 +1,321 @@
...
+            // Show any errors
+            for (Throwable error : errors) {
+                System.out.println("\n### ERROR ###\n");
+                if (error instanceof CloudException) {
+                    CloudException ce = (CloudException) error;
+                    System.out.println("Cloud Exception: " + ce.getMessage());
+                } else {
+                    error.printStackTrace();
+                }
+            }
```
anuchandy: Due to the reasons mentioned in the previous comment, we could use single error variable instead of list and replace the for loop with:

```java
if (error != null) {
    if (error instanceof rx.CompositeException) {
       for (Throwable t : ((rx.CompositeException)(error)).getExceptions()) {
           if (error instanceof CloudException) {
               CloudException ce = (CloudException) t;
                System.out.println("Cloud Exception: " + ce.getMessage())
           } else {
                t.printStackTrace();
           }
       }
   } else if (error instanceof CloudException) {
       CloudException ce = (CloudException) error;
       System.out.println("Cloud Exception: " + ce.getMessage())
   } else {
       error.printStackTrace();
   }
}
```

azure-samples/src/main/java/com/microsoft/azure/management/compute/samples/CreateVirtualMachinesAsyncTrackingRelatedResources.java
anuchandy: Sorry, we still need onErrorReturn due to last() operator

azure-samples/src/main/java/com/microsoft/azure/management/compute/samples/CreateVirtualMachinesAsyncTrackingRelatedResources.java
martinsawicki: Today, I understand that we bail out of the parallel creation as soon as we hit the first error. But in the past we talked about being more eager about that and continuing creating VMs, and just accumulate errors (`mergeDelayError()` semantics, rather than `merge()`, or something like that).  Wouldn't the current logic better protect against that potential future implementation?

anuchandy: hmm, the current logic not really protect against outcome of potential future implementation where we continue creating resources eagerly. The reason is - in both cases, the we will be conforming to Observable rule where the terminal events `onError` or `onComplete` must invoked only once. Due to this rule, if there are multiple error happens (either due to parallel errors in case of `merge` or due to delaying of error due to `mergeDelayError`) internally they will be aggregated and thrown as Composite exception. So in nutshell, an observable will never throw more than once.

martinsawicki:
@anuchandy if you could take a quick second look, that'd be appreciated. B/c of now using `.subscribe()` instead of `.map()`, some stuff had to be rearranged.  It seems to work fine still.

anuchandy:
@martinsawicki LGTM!

PR #50, opened 2017-11-14
Container service updates

milismsft:
Mostly Csharp related conversion and update notes.


azure-mgmt-containerservice/src/main/java/com/microsoft/azure/management/containerservice/ContainerService.java
```diff
@@ -33,71 +34,85 @@
     /**
      * @return the master node count
      */
+    @Method
```
martinsawicki: why are these all marked as metods - do they make an actual separate call to azure?  If so, then they should be prefixed wuth "get". But from the look of them, they look like what should be properties on the C# side, so not sure why `@Method` is on them.

PR #51, opened 2017-11-14
Add sample for delegated credentials

jianghaolu:


jianghaolu:
Pending https://github.com/Azure/autorest-clientruntime-for-java/pull/298

PR #63, opened 2017-11-18
[Do not merge] Batch AI

lenala:


azure-mgmt-batchai/src/main/java/com/microsoft/azure/management/batchai/BatchAICluster.java
```diff
@@ -0,0 +1,253 @@
...
+         */
+        interface WithVMSize {
+            /**
+             * @param vmSize virtual machine size
+             * @return next stage of the definition
+             */
+            WithUserName withVMSize(String vmSize);
+        }
+
+        interface WithUserName {
```
jianghaolu: Add javadocs

azure-mgmt-batchai/src/main/java/com/microsoft/azure/management/batchai/ToolTypeSettings.java
```diff
@@ -0,0 +1,260 @@
...
+                WithAttach<ParentT> withWorkerCount(int workerCount);
+
+                WithAttach<ParentT> withParameterServerCount(int parameterServerCount);
+            }
+
+            interface Blank<ParentT> extends WithPython<ParentT> {
+            }
+
+            interface WithPython<ParentT> {
+                WithMasterCommandLineArgs<ParentT> withPython(String pythonScriptFilePath);
```
jianghaolu: `withPythonScriptFile()`?

azure/src/main/java/com/microsoft/azure/management/Azure.java
```diff
@@ -666,6 +671,20 @@ public BatchAccounts batchAccounts() {
...
+     */
+    public BatchAIClusters batchAIClusters() {
+        return batchAIManager.clusters();
+    }
+
+    /**
+     * @return entry point to managing batch AI file servers.
+     */
+    public FileServers fileServers() {
+        return batchAIManager.fileServers();
```
anuchandy: The name  `fileServers()` is very generic. Since this is roll up client, can we be more specific here like `batchAIFileServers()` or some name which avoid confusion when other RP also uses the term `FileServers` in future.

azure/pom.xml
```diff
@@ -193,6 +193,11 @@
       <artifactId>commons-io</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.microsoft.azure</groupId>
+      <artifactId>azure-mgmt-batchai</artifactId>
+      <version>1.5.2-beta-SNAPSHOT</version>
```
anuchandy: We usually version module as `beta` only if we are going to publish just the generated code, here we have fluent interface for batch ai service so we should version it as `1.5.2-SNAPSHOT`

PR #73, opened 2017-11-29
App Gateway URL path map support

martinsawicki:
dependent on `@Beta(SinceVersion.V1_5_0)`


anuchandy:
@martinsawicki just FYI adding don't merge for this, just to make sure our repo is ready in case we want to do a release before scheduled jan release.

PR #85, opened 2017-12-08
Re-gen Azure Container Instance inners and implement restart policy

milismsft:


milismsft:
There's one particular change I'd like to be reviewed, that is overwriting TopLevelModifiableResourcesImpl list methods. The Azure ACI service guys claim that they cannot return a full resource body because it's very expensive for them but the SDK user definitely expects all the information as part of the API return. So I had to work around it in the SDK to "refresh" the resources in the list in order to guarantee they have all the properties set.

anuchandy:
@milismsft I'll take a look. Just FYI marking this as `DON'T MERGE` as we talked yesterday, because we might be doing a release before Jan, so trying to keep our repo ready for that.

milismsft:
I'm actually fine with merging these changes since they primarily address bug fixes and user requested features. The delta between these updates and existing code is minor and the risk is very small.

PR #103, opened 2017-12-14
Update README for 1.5.1

anuchandy:


anuchandy:
1.2.1 autorest runtime for Java needs to be published with MSICredentials for EMSI and readme's snapshot section should be updated to use 1.2.2-SNAPSHOT of runtime.

PR #107, opened 2017-12-17
Fixing the issue of not invoking post run tasks added via beforeGroupInvoke of root task

anuchandy:
This PR fixes the issue of NOT invoking “post run dependents” added via `beforeGroupInvoke` method of ROOT-task. The existing logic already handles the “post run dependents” added through `beforeGroupInvoke` methods in NON-ROOT tasks.


anuchandy:
@jianghaolu, this fixes the issue you encountered while moving app-settings and other post run activities to `beforeGroupCreateOrUpdate()` of AppServiceImpl. 

Give it a try again once this PR merged. Note that this will work in UPDATE as well, because `beforeGroupCreateOrUpdate()` will be invoked during update, which cause adding NEW post-run activities that graph runs once subscribed [and the previously activities added during CREATE will be ignored].

PR #109, opened 2017-12-21
Key Vault data plane bridge

jianghaolu:


azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Key.java
```diff
@@ -0,0 +1,457 @@
...
+         * The stage of a key definition allowing to specify whether to store the key in
+         * hardware security modules.
+         */
+        interface WithHsm {
+            /**
+             * Specifies whether to store the key in hardware security modules.
+             * @param isHsm store in Hsm if true
+             * @return the next stage of the definition
+             */
+            WithImport withHsm(boolean isHsm);
```
milismsft: What about two methods here, withHsm() and without(Hsm)? Or default for create to whatever the default is (I will guess isHsm == false which translate into one method with(Hsm)) and only have the two as part of the update def.

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Key.java
```diff
@@ -0,0 +1,457 @@
...
+         * The stage of a key update allowing to specify whether to store the key in
+         * hardware security modules.
+         */
+        interface WithHsm {
+            /**
+             * Specifies whether to store the key in hardware security modules.
+             * @param isHsm store in Hsm if true
+             * @return the next stage of the update
+             */
+            UpdateWithImport withHsm(boolean isHsm);
```
milismsft: See my comment from above...

PR #111, opened 2017-12-28
App Service Implicit MSI

jianghaolu:


azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/WebAppBase.java
```diff
@@ -216,6 +216,12 @@
      */
     String autoSwapSlotName();
 
+    /**
+     * @return the managed service identity in Azure Active Directory for this web app
+     */
+    @Beta(SinceVersion.V1_6_0)
+    ManagedServiceIdentity managedServiceIdentity();
```
anuchandy: One suggestion here is: It would be nice to be more explicit here. Instead of returning `ManagedServiceIdentity` we could have two separate properties 

`string systemAssignedManagedServiceIdentityTenantId()`, 
`string systemAssignedManagedServiceIdentityPrincipalId()`. 

This will guard us from making breaking change when web app eventually supports `user assigned` managed service identity.  Also ensures consistency in MSI interface across fluent resources (VM, VMSS, Storage).

Btw just FYI we also enabled `isManagedServiceIdentityEnabled()` getter in other resources (VM, VMSS, Storage) that supports MSI

anuchandy: Also in the gen-code there is this property `object ManagedServiceIdentity::type` of type object. I think we should ask web-app team to be specific here in swagger, today the supported value for type in case of other RPs are `SystemAssigned`, `UserAssigned`, `SystemAssigned,UserAssigned` and `None`.

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/WebAppBase.java
```diff
@@ -1229,6 +1250,20 @@
...
+         * @param <FluentT> the type of the resource
+         */
+        @Beta(SinceVersion.V1_6_0)
+        interface WithManagedServiceIdentity<FluentT> {
+            /**
+             * Specifies that System Assigned Managed Service Identity needs to be enabled in the web app.
+             * @return the next stage of the web app definition
+             */
+            @Method
+            WithCreate<FluentT> withSystemAssignedManagedServiceIdentity();
```
anuchandy: you mean return type `Update<FluentT>`? Same for `withContainerLoggingDisabled()` return type

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/WebAppBase.java
jianghaolu: Yes the Object is actually just a string. I'll look at those resources and try to make it consistent.

PR #125, opened 2018-01-19
Batch AI - minor updates

lenala:


azure-mgmt-batchai/src/main/java/com/microsoft/azure/management/batchai/BatchAIFileServer.java
```diff
@@ -24,6 +25,48 @@
         GroupableResource<BatchAIManager, FileServerInner>,
         Refreshable<BatchAIFileServer> {
 
+    /**
+     * @return the size of the virtual machine of the File Server.
+     * For information about available VM sizes for File Server from the
+     * Virtual Machines Marketplace, see Sizes for Virtual Machines (Linux).
+     */
+    String vmSize();
```
anuchandy: Do we need the text "For information about available VM sizes for File Server from the Virtual Machines Marketplace, see Sizes for Virtual Machines (Linux).", not sure it is clear enough.

PR #142, opened 2018-01-25
Removing unneeded test for Batch AI

lenala:


anuchandy:
@lenala will merge as soon as CI is green, thanks!

PR #162, opened 2018-02-05
Use gulpfile-specified output folder

RikkiGibson:
Fixes a breaking config change that was introduced in azure-rest-api-specs.

It seems to me that while AutoRest config which affects the meaning of the generated code, like `input-file` and `title` may belong in azure-rest-api-specs, but stuff like `output-folder` and `clear-output-folder` is project-specific and shouldn't necessarily come from azure-rest-api-specs.

I'm not invested in this working any particular way so feel free to make any changes to the gulpfile in here @daschult as long as the code generation works :)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -121,7 +121,7 @@ var codegen = function(project, cb) {
                         ' --azure-arm ' +
                         generator +
                         ` --namespace=${mappings[project].package} ` +
-                        ` --output-folder=${outDir} ` +
+                        ` --java.output-folder=${outDir} ` +
                         ` --license-header=MICROSOFT_MIT_NO_CODEGEN ` +
                         generatorPath +
                         regenManager +


gulpfile.js
```diff
@@ -121,7 +121,7 @@ var codegen = function(project, cb) {
                         ' --azure-arm ' +
                         generator +
                         ` --namespace=${mappings[project].package} ` +
-                        ` --output-folder=${outDir} ` +
```
milismsft: I had a slightly different fix for this: https://github.com/Azure/azure-libraries-for-java/pull/159/files#diff-b9e12334e9eafd8341a6107dd98510c9L123

RikkiGibson:
OK, I'm finding this fix might not actually be working-- I was certain I had it working seconds ago. Very strange. Might wind up opening another PR.

RikkiGibson:
It seems that in the time since I made this fix, azure-rest-api-specs changed the name of the branch we want from "current" to "master", so AutoRest just always deleted everything. I'll open another PR with that fix and update the .NET PR.

PR #168, opened 2018-02-08
Java SDK For new version 2018-01-31

kjeur:
updating sdk by https://github.com/Azure/azure-rest-api-specs/blob/master/specification/consumption/resource-manager/readme.md#java


milismsft:
Why are we getting rid of the Java code and replace it with Python??? Perhaps this PR was submitted to the wrong repo?

kjeur:
@milismsft My bad, cancelling PR.

PR #169, opened 2018-02-08
[Don't merge] Java SDK For new version 2018-01-31

kjeur:
Updating SDK for new verssion 2018-01-31

Reference
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/consumption/resource-manager/readme.md#java

PR #170, opened 2018-02-09
Consumption 2018 01

kjeur:
SDK For new version 2018-01-31

Note : https://github.com/Azure/azure-libraries-for-java/pull/169 - creating new one with suggested commit messages


milismsft:
@kjeur I've added the "do not merge" label for now since we are not completely done with releasing the 1.6 version of the Java SDK. We should be able to merge this in the next couple days though.

BTW let me know if you will be interested in having some minimal Fluent layer on top of the service rather than working directly with the inner classes (which is not the supported way).

kjeur:
@milismsft please merge once you are ready with your release version.
@ms-premp to answer Fluent layer question.

ms-premp:
@milismsft  we will be interested in Fluent layer.. is it auto-generated ? If yes, then please let us know, what action item we have for doing it.
If there is some significant work involved on our side to get on Fluent, then right now I will opt for non-fluent code. We are running tight on our release schedules so cannot space out for this task at present.

milismsft:
The Fluent layer can not be fully autogenerated at the moment unfortunately. It also looks like the "consumption" Swagger spec is a bit more special than most of other services which complicates the things further (i.e. two distinct createOrUpdate for a Budget). For now I'm considering adding a generic Fluent layer with the possibility to refine it further in the future; IMHO still better than directing the users to call through the "inner" methods. I'll keep you posted if that can go in into the next release.

ms-premp:
Thanks a ton @milismsft

PR #176, opened 2018-02-13
Event hub fluent implementation

anuchandy:


anuchandy:
Merging this PR.. I will be sending another PR with updated child definitions that follows a strict pattern

```java
  parents()
     .children()
        .define(..)
        .withExistingParent(Parent p) | withExistingParent(string, string...) | withExistingParentId(string id)
        .<any-withers-relevant-to-child>
        .create();
```

PR #187, opened 2018-02-20
[DoNotMerge] CDN experiment

lmazuel:
@daschult That's the result of the SwaggerToSdk conf we have based on latest CDN in this PR:
https://github.com/Azure/azure-rest-api-specs/pull/2514

I used the SwaggerToSdk CLI:
```shell
generate_sdk.exe -v -m ..\azure-rest-api-specs\specification\cdn\resource-manager\readme.md
```
Could you check if it's fair, meaning diff can be explained ONLY by:
- Newer Autorest
- Newer Swagger

If you see something obviously wrong that comes from a configuration issue, please reach out to me. The plan is to fix the PR 2514 until this Java is is a good shape. When we reach that, we know the system is ready (whatever the real planning).

Thanks!

FYI @salameer


daschult:
Hi @lmazuel. Sorry it took me so long to get back to you. I looked over this and everything looks good on our side. I had to make a [fix ](https://github.com/Azure/autorest.java/pull/164) in the autorest.java generator (missing import statement), but once that change gets merged and published everything should work fine.

daschult:
Everything looks good. I'm going to close (not merge) this PR.

PR #194, opened 2018-02-22
[Event Hub]: Updating child defintion to follow new consistent pattern

anuchandy:


anuchandy:
Based on the comment https://github.com/Azure/azure-libraries-for-java/pull/176#issuecomment-367549620

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
```diff
@@ -72,11 +72,11 @@ public void canManageEventHubNamespaceBasicSettings() {
...
                 .define(namespaceName2)
-                .withRegion(REGION)
-                .withExistingResourceGroup(RG_NAME)
-                // SDK should use Sku as 'Standard' and set capacity.capacity in it as 11
-                .withCurrentThroughputUnits(11)
-                .create();
+                    .withRegion(REGION)
+                    .withExistingResourceGroup(RG_NAME)
+                    // SDK should use Sku as 'Standard' and set capacity.capacity in it as 11
+                    .withCurrentThroughputUnits(11)
```
milismsft: Might be worth simplifying this to .withThroughputUnits() since "current" is redundant in this context. Does this method set a limit or a fixed capacity?

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
```diff
@@ -88,10 +88,10 @@ public void canManageEventHubNamespaceBasicSettings() {
 
         EventHubNamespace namespace3 = eventHubManager.namespaces()
                 .define(namespaceName3)
-                .withRegion(REGION)
-                .withExistingResourceGroup(RG_NAME)
-                .withSku(EventHubNamespaceSkuType.BASIC)
-                .create();
+                    .withRegion(REGION)
+                    .withExistingResourceGroup(RG_NAME)
+                    .withSku(EventHubNamespaceSkuType.BASIC)
```
milismsft: In SQL for databases and elastic pools I simplified this further: .withBasicSku() or .withStandardSku() etc.
It makes sense if the event hub SKU selection is small enough. 
Is it possible to combine the SKU selection with the throughput unit setting? If yes, you could combine the two like:
.withStandardSku(int capacity)... Just an idea :-)

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
```diff
@@ -149,11 +149,11 @@ public void canManageEventHubNamespaceEventHubs() throws Exception {
 
         eventHubManager.namespaces()
                 .eventHubs()
-                .define(eventHubName3)
-                .withExistingNamespaceId(namespace.id())
-                .withPartitionCount(5)
-                .withRetentionPeriodInDays(6)
-                .create();
+                    .define(eventHubName3)
+                    .withExistingNamespaceId(namespace.id())
```
milismsft: 👍

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
```diff
@@ -225,15 +225,15 @@ public void canManageEventHubConsumerGroups() throws Exception {
...
+                    .withRegion(REGION)
+                    .withNewResourceGroup(RG_NAME);
 
         EventHub eventHub = eventHubManager.eventHubs()
                 .define(eventHubName)
-                .withNewNamespace(namespaceCreatable)
-                .withNewConsumerGroup("grp1")
-                .withNewConsumerGroup("grp2", "metadata111")
-                .create();
+                    .withNewNamespace(namespaceCreatable)
```
milismsft: Are you creating a "parent" namespace as part of the "child" eventHub definition? That will be wrong...

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
```diff
@@ -342,27 +342,27 @@ public void canConfigureEventHubDataCapturing() throws Exception {
...
-                .withNewNamespace(namespaceCreatable)
-                .withNewStorageAccountForCapturedData(storageAccountCreatable, containerName1)
-                .withDataCaptureEnabled()
-                // Window config is optional if not set service will choose default for it2
-                //
-                .withDataCaptureWindowSizeInSeconds(120)
-                .withDataCaptureWindowSizeInMB(300)
-                .create();
+                    .withNewNamespace(namespaceCreatable)
+                    .withNewStorageAccountForCapturedData(storageAccountCreatable, containerName1)
```
milismsft: We might need a better name for this method, something like .withDataCapturingContainer() may be?

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
```diff
@@ -342,27 +342,27 @@ public void canConfigureEventHubDataCapturing() throws Exception {
...
+                    .withDataCaptureEnabled()
```
milismsft: this can be default/implicit and only provide an overwrite, .withDataCaptureDisable() - update will have them both though. I had a similar pattern for a particular SQL firewall rule (withAccessFromAzureServices())...

azure-mgmt-eventhub/src/main/java/com/microsoft/azure/management/eventhub/AuthorizationRule.java
```diff
@@ -77,7 +77,7 @@
              * @return the next stage of the definition
              */
             @Beta(Beta.SinceVersion.V1_7_0)
-            T withListeningEnabled();
+            T withListenAccess();
```
milismsft: 👍

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
anuchandy: There are two ThroughputUnit config here, the throughput currently set for the event hub and the max throughput service is allowed to expand to, in-order to distinguish between these two we made it explicit in the method names - `withCurrentThroughputUnits` and `withThroughputUnitsUpperLimit`

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
anuchandy: This was an ask from event-hub team given that users are more likely to directly start creating event-hub. My question here in general is - do we need to support this if childResource accessor in the current context takes more priority over parent?

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
anuchandy: This cannot be default because enabling this add additional cost and any thing that cause unintentional cost should be explicit in my opinion.

milismsft: but the intention is already there since it's the user's code to specify the data capturing properties. It's like creating a virtual machine and you will ask the user to specifically set the state of the machine as powered on.

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
milismsft: sounds good.

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
milismsft: In this case you already pass a Creatable which means that the user already goes through the process of specifying the required parameters to create the namespace. The user might just call "create()" after that to complete this creation process rather than going through the child resource creation. Also it's not like the two operation can be run in parallel since the event hub creation requires the namespace to be created first.

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
milismsft: you could overload the .withBasicSku() to take one (the upper limit I will guess) or both as parameters. In SqlDatabases I have a similar pattern where user can specify the edition, the service objective and the max storage capacity.

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
anuchandy: Thinking more about it, I agree with you. Another issue is though we can do something like this for level1 children, it is hard to do the same for level2 onwards. Let's simplify this and provide only one option which is doable in all levels. I will remove it. Thanks!

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
anuchandy: Yes agree there is a different ways to express it but as I mentioned my intention was to make it clear that these two totally differnt properties in the payload are logically related, hence used matchng names. In addition this is how eventhub wanted to express if we refer their docs.

azure-mgmt-eventhub/src/test/java/com/microsoft/azure/management/eventhub/EventHubTests.java
anuchandy: What I tried to provide is,  allow user to configure them but not necessarily enable it at the time of creation and whenever their app want to capture events enable it and once they have enough data to analyze it then disable it and so on. Irrespective of you turn it on or off event hub continue to work.

This is different from vm example you described right? Vm is a top level resource and it is implicit that it will be automatically turned on upon creation same goes with event hub as soon as it is created it is live for consumption. When we create such top-level or use a Creatable it is known to incur cost. Eventhub data capture is an additional config and which you can setup but setup won't cost you money but soon you decide to start using it you pay for it and you can stop paying it by Turing it off separately but not Turing off the parent resource.

PR #196, opened 2018-02-22
Update for SQL implementation, tests and recordings

milismsft:


anuchandy:
The patterns followed in the SQL has been discussed in detail for last couple of months (during development stage) and is looks good.

PR #229, opened 2018-03-06
Fixed bug where blank slot configurations and updates would not apply app settings and connection strings

npeuser:
After some consideration and more investigation into how the Azure SDK components work, I came up with this solution, which fixes both issues and seems to be the original intent of the check for webAppBase == null || !isInCreateMode - apply the current configuration instead of reading it from another source, if there is no base to create from or if the slot is not being based on an existing configuration element.


jianghaolu:
LGTM. Thanks!!

PR #235, opened 2018-03-08
SQL bug fixes - withTags() does not return the correct stage

milismsft:
Minor fixes when specifying withTags() for SQL databases and elastic pools.


hovsepm:
@milismsft Please fix tag assignment everywhere.... in the solar system... :)

PR #237, opened 2018-03-08
Fix for throttling interceptor when multiple subscriptions are used.

milismsft:
Added a concurrent map instance to manage lock per each subscription.

### DIFF
--- a/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ResourceManagerThrottlingInterceptor.java
+++ b/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ResourceManagerThrottlingInterceptor.java
@@ -29,7 +29,6 @@ import java.util.regex.Pattern;
  */
 public class ResourceManagerThrottlingInterceptor implements Interceptor {
     private static final String LOGGING_HEADER = "x-ms-logging-context";
-    private static final ReentrantLock REENTRANT_LOCK = new ReentrantLock();
     private static final ConcurrentMap<String, ReentrantLock> REENTRANT_LOCK_MAP = new ConcurrentHashMap<>();
 
     @Override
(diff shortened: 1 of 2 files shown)


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ResourceManagerThrottlingInterceptor.java
```diff
@@ -27,14 +30,17 @@
 public class ResourceManagerThrottlingInterceptor implements Interceptor {
     private static final String LOGGING_HEADER = "x-ms-logging-context";
     private static final ReentrantLock REENTRANT_LOCK = new ReentrantLock();
```
RikkiGibson: remove

azure-mgmt-sql/src/test/java/com/microsoft/azure/management/sql/SqlServerOperationsTests.java
```diff
@@ -33,6 +33,30 @@
     private static final String START_IPADDRESS = "10.102.1.10";
     private static final String END_IPADDRESS = "10.102.1.12";
 
+    @Test
```
RikkiGibson: remove

azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ResourceManagerThrottlingInterceptor.java
```diff
@@ -26,15 +29,17 @@
...
     @Override
     public Response intercept(Chain chain) throws IOException {
         // Gate control
+        String subscriptionId = ResourceUtils.subscriptionFromResourceId(chain.request().url().url().getPath());
+        REENTRANT_LOCK_MAP.putIfAbsent(subscriptionId, new ReentrantLock());
         try {
-            synchronized (REENTRANT_LOCK) {
-                if (REENTRANT_LOCK.isLocked()) {
-                    REENTRANT_LOCK.wait();
+            synchronized (REENTRANT_LOCK_MAP.get(subscriptionId)) {
```
RikkiGibson: It looks like this is synchronizing on both the intrinsic lock and the ReentrantLock methods, which seems off. But if it works, it works.

PR #385, opened 2018-03-24
SQL: renaming...

milismsft:
Small Fluent design change removing the "plural" from the method name...


milismsft:
@jianghaolu can you merge this unless it's too late for this release?

PR #410, opened 2018-04-10
Fix JavaDoc NPE exception when generating JavaDoc under JDK 9

JonathanGiles:
Update JavaDoc comment that incorrectly attempts to use @link to reference a type U, where U is not a class that can be looked up. This results in a JavaDoc NPE when using JDK 9, preventing JavaDoc from being generated for this library due to the issue discussed at https://bugs.openjdk.java.net/browse/JDK-8188248


JonathanGiles:
Thanks!

PR #414, opened 2018-04-11
Updated DNS with the API version "2018-03-preview" from "2016-04"

hovsepm:


azure-mgmt-dns/src/main/java/com/microsoft/azure/management/dns/implementation/ARecordSetImpl.java
```diff
@@ -21,15 +21,13 @@
 class ARecordSetImpl
         extends DnsRecordSetImpl
         implements ARecordSet {
-    ARecordSetImpl(final DnsZoneImpl parent, final RecordSetInner innerModel) {
-        super(parent, innerModel);
+    ARecordSetImpl(final String name, final String type, final DnsZoneImpl parent, final RecordSetInner innerModel) {
```
anuchandy: Do we need the second param `final String type` for each specialized record-set?, given each impl know (e.g. ARecordSetImpl) what type it is. I mean, was thing like

```java
ARecordSetImpl(final String name, final DnsZoneImpl parent, final RecordSetInner innerModel) { 
    super(name, RecordType.A.toString(), parent, innerModel); 
}
```

azure-mgmt-dns/src/main/java/com/microsoft/azure/management/dns/implementation/DnsZoneImpl.java
```diff
@@ -414,31 +479,67 @@ private void initRecordSets() {
...
+        this.inner().withRegistrationVirtualNetworks(null);
+        this.inner().withResolutionVirtualNetworks(null);
+        return this;
+    }
+
+    @Override
+    public DnsZoneImpl withPrivateAccess() {
+        this.inner().withZoneType(ZoneType.PRIVATE);
+        this.inner().withRegistrationVirtualNetworks(null);
+        this.inner().withResolutionVirtualNetworks(null);
```
anuchandy: Is it safe to reset the reg and resolution networks? let say user is updating an existing private zone, if he choose `withPrivateAccess` then that will clear all existing reg and resolution networks without user knowledge.

anuchandy:
Merge this after this PR https://github.com/Azure/azure-libraries-for-java/pull/416 is merged, it add snapshot repo ref to pom file, which is needed for snapshot runtime bits.

PR #430, opened 2018-04-22
Updating resource module to use fluentcore package in "azure java runtime"

anuchandy:
Removing types in "azure java lib" `fluentcore` package those are moved to "azure java runtime" and instead uses them from the runtime. PR also refactor some existing types to use new base types in runtime. See this [runtime PR](https://github.com/Azure/autorest-clientruntime-for-java/pull/426) for more details.


milismsft:
@anuchandy Per our short conversation, one other option is to move these common dependencies into their own JAR/library rather than the Runtime. This will allow us to keep the flexibility on making changes to this code without worrying about shipping an updated Runtime first. Just an idea...

anuchandy:
@milismsft - the fluentcore classes we moved to runtime are pretty stable, the types such as ARM `Resource`, `ProxyResource` are already runtime. So I thought fluentcore types like `GroupableResouce` are better stay with these ARM resources. 

The DAG types can be a separate jar but they are also stable enough. We don't want to maintain DAG only as a general purpose lib for job scheduling, there are popular libs out there for that.  I'm lean more towards maintaining one less dependency. 

Also often make updates to runtime (see the commits), if there is another jar that takes runtime dependency then we need to update dependency in that  jar as well otherwise we end up loading different runtime versions in fluent lib. I was thinking of starting with this approach and if we see issues in the next couple of releases then we can brainstorm  modularizing it further.

anuchandy:
PR not valid anymore

PR #431, opened 2018-04-24
CosmosDB: Add missing implementations for creating other data models

milismsft:
Add missing implementations to create Cassandra, Azure Table and Gremlin data model CosmosDB accounts.

### DIFF
--- a/azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/CosmosDBAccount.java
+++ b/azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/CosmosDBAccount.java
@@ -8,6 +8,7 @@ package com.microsoft.azure.management.cosmosdb;
 import com.microsoft.azure.management.apigeneration.Beta;
 import com.microsoft.azure.management.apigeneration.Fluent;
 import com.microsoft.azure.management.apigeneration.Beta.SinceVersion;
+import com.microsoft.azure.management.apigeneration.Method;
 import com.microsoft.azure.management.cosmosdb.implementation.CosmosDBManager;
 import com.microsoft.azure.management.cosmosdb.implementation.DatabaseAccountInner;
 import com.microsoft.azure.management.resources.fluentcore.arm.Region;
@@ -102,6 +103,13 @@ public interface CosmosDBAccount extends
     Observable<DatabaseAccountListConnectionStringsResult> listConnectionStringsAsync();
 
     /**
+     * @return a list that contains the Cosmos DB capabilities
+     */
+    @Beta(SinceVersion.V1_10_0)
+    List<Capability> capabilities();
+
+
+    /**
      * @param keyKind the key kind
      */
     void regenerateKey(KeyKind keyKind);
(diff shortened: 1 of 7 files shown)


azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/CosmosDBAccount.java
```diff
@@ -154,6 +162,61 @@
...
+            WithConsistencyPolicy withDataModelCassandra();
+
+            /**
+             * Creates an Azure Table CosmosDB account.
+             *
+             * @return the next stage of the definition
+             */
+            @Method
+            @Beta(SinceVersion.V1_10_0)
+            WithConsistencyPolicy withDataModelAzureTable();
```
hovsepm: >withDataModelAzureTable [](start = 34, length = 23)

Should it be just withDataModel method and then a parameter of Enum type with all the supported DB Types and enum members? Too many methods in Intelisence that start with the same prefix (A long one in this case) creates a bunch of clutter IMHO

hovsepm: also it is kind of weird construct `withDataModelSql` . More natural would be `withSqlAccount` or `withSqlTyoe` but in that case you will have types in a hard to find order in the intelisence. So reiterating my proposal to have one method and bunch of enum values :)

JREndeanMSFT: In the management portal it is called API 
cosmosdbportalui

azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/implementation/CosmosDBAccountImpl.java
```diff
@@ -162,6 +173,52 @@ public CosmosDBAccountImpl withKind(DatabaseAccountKind kind) {
...
+    public CosmosDBAccountImpl withKind(DatabaseAccountKind kind, Capability... capabilities) {
+        this.inner().withKind(kind);
+        this.inner().withCapabilities(Arrays.asList(capabilities));
+        return this;
+    }
+
+    @Override
+    public CosmosDBAccountImpl withDataModelSql() {
+        this.inner().withKind(DatabaseAccountKind.GLOBAL_DOCUMENT_DB);
+        return this;
```
JREndeanMSFT: Are you sure the "defaultExperience" tags are not needed on these? If you create each API type in the portal, they all have a "defaultExperience" tag.

milismsft: The tags do not impact resource provisioning; if they do then it's a bug. While the tests I've added set a generic tag it's not required to create an CosmosDB account.

azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/CosmosDBAccount.java
milismsft: We do not need to match exactly what the Azure Portal does. Data model fits better the type of the account (see the explanation behind the little "i").

There's already a method that takes enums (see the withKind() overload I just added). These new methods are direct shortcuts for creating the respective CosmosDB account.

PR #432, opened 2018-04-25
Added enableSoftDelete, DeletedVault, and Purge to method group

tiffanyachen:


azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/implementation/VaultsImpl.java
```diff
@@ -97,4 +101,42 @@ protected VaultImpl wrapModel(VaultInner vaultInner) {
...
+	private PagedList<DeletedVault> wrapPage(PagedList<DeletedVaultInner> listDeleted) {
+		PagedListConverter<DeletedVaultInner, DeletedVault> converter =  
+				new PagedListConverter<DeletedVaultInner, DeletedVault> () {
+					@Override
+					public Observable<DeletedVault> typeConvertAsync(DeletedVaultInner inner) {
+						return Observable.just((DeletedVault) new DeletedVaultImpl(inner));
+					}
+		};
+		return converter.convert(listDeleted);
+	}
```
anuchandy: This is correct.

Just make sure we are not using the name `wrapPage`, we could move converter variable def to `public PagedList<DeletedVault> listDeleted()` method.

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vaults.java
```diff
@@ -28,23 +37,260 @@
...
+    /**
+     * Create or update a key vault in the specified subscription.
+     *
+     * @param resourceGroupName The name of the Resource Group to which the server belongs.
+     * @param vaultName Name of the vault
+     * @param parameters Parameters to create or update the vault
+     * @throws IllegalArgumentException thrown if parameters fail the validation
+     * @return the observable for the request
+     */
+    public Observable<ServiceResponse<Vault>> createOrUpdateWithServiceResponseAsync(String resourceGroupName, String vaultName, VaultCreateOrUpdateParameters parameters);
```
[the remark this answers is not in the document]
tiffanyachen: What about async methods that take a callback and are exposed by VaultsInner?

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
```diff
@@ -87,6 +87,24 @@
      * retrieve secrets from the key vault.
      */
     boolean enabledForTemplateDeployment();
+    
+    /**
+     * @return whether soft delete is enabled for this key vault.
+     */
+    boolean softDeleteEnabled();
```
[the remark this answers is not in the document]
milismsft: @tiffanyachen in this case you need to be consistent with the rest of the Java SDK naming patterns for this service and not other languages. There's a good reason and lots of review discussions before we chose the naming we use today. So unless you want to change all the other method names and introduce your own new naming pattern, then work around the backward compatibility breaks resulted from these change, then my recommendation is to stay with @anuchandy feedback.

milismsft:
@tiffanyachen I have a similar change that only addresses enabling soft delete and purge protection at https://github.com/milismsft/azure-libraries-for-java/commit/b52185bc43ce9c4d7ceecc2f02e6e1dfc6308a7e. Would you be able to incorporate it into your changes?

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/CheckNameAvailabilityResult.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.azure.management.keyvault;
+
+import com.microsoft.azure.management.apigeneration.Fluent;
+import com.microsoft.azure.management.keyvault.implementation.CheckNameAvailabilityResultInner;
+import com.microsoft.azure.management.resources.fluentcore.model.HasInner;
+
+/**
+ * Wrapper for CheckNameAvailabilityResultInner
+ * @author tifchen
```
milismsft: Can you please remove this comment and change the overall description of the method to something like "The result of checking for the Key Vault name availability."?

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/CheckNameAvailabilityResult.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Wrapper for CheckNameAvailabilityResultInner
+ * @author tifchen
+ *
+ */
+@Fluent(ContainerName = "/Microsoft.Azure.Management.Fluent.KeyVault")
+public interface CheckNameAvailabilityResult extends
+    HasInner<CheckNameAvailabilityResultInner>{
+
```
milismsft: I would expect to see here some methods such as isAvailable(), unavailabilityReason() and unavailabilityMessage()...

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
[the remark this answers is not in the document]
schaabs: @anuchandy @milismsft In the case of the first three fields, enableForDeployment, enabledForDiskEncryption, and enabledForTemplateDeployment, they are named this way because the user is enabling the vault to be accesed **FOR** arm deployments, disk encryption, and template deployment.  In the case of softDeleteEnabled and purgeProtectionEnabled the user is enabling the features soft delete and purge protection on the vault.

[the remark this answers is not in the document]
milismsft: @schaabs I don't understand how is soft delete any different then the others... It looks like it is required for instance when setting up SQL TDE (see customer report https://github.com/Azure/azure-libraries-for-net/issues/315).
If "soft delete" and "purge protection" are simple vault properties it might be worth removing the whole "enabled" thing from the name, i.e "bool softDelete()" and "bool purgeProtection()".

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/CheckNameAvailabilityResult.java
tiffanyachen: Added in latest commit

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
```diff
@@ -204,6 +236,13 @@
              * @return the next stage of key vault definition
              */
             WithCreate withTemplateDeploymentDisabled();
+            
+            /**
+             * Set the createMode value. 
+             * 
+             * @return the next stage of key vault definition
+             */
+            WithCreate withCreateMode(CreateMode createMode);
```
milismsft: @tiffanyachen Could CreateMode.RECOVER be used with an existing Key Vault resource that has not been deleted? What impact other settings prior to reaching this method during the create flow will have when this value is set?

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/implementation/VaultsImpl.java
```diff
@@ -91,10 +110,250 @@ protected VaultImpl wrapModel(VaultInner vaultInner) {
...
+    }    
+	
+	@Override
+	public PagedList<DeletedVault> listDeleted() {
+	    PagedList<DeletedVaultInner> listDeleted = this.inner().listDeleted();
+	    PagedListConverter<DeletedVaultInner, DeletedVault> converter =  
+	            new PagedListConverter<DeletedVaultInner, DeletedVault> () {
+	                @Override
+                    public Observable<DeletedVault> typeConvertAsync(DeletedVaultInner inner) {
+                        return Observable.just((DeletedVault) new DeletedVaultImpl(inner));
```
milismsft: @tiffanyachen the cast is not needed here (if DeletedVaultImpl class implements DeletedVault interface).

milismsft:
@tiffanyachen the Travis build is failing because of check-style errors; you can try to reproduce them locally by executing "mvn checkstyle:check".

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/VaultAccessPolicyParameters.java
```diff
@@ -0,0 +1,13 @@
...
+import com.microsoft.azure.management.keyvault.implementation.VaultAccessPolicyParametersInner;
+import com.microsoft.azure.management.resources.fluentcore.model.HasInner;
+
+/**
+ * Parameters for updating the access policy in a vault.
+ */
+@Fluent(ContainerName = "/Microsoft.Azure.Management.Fluent.KeyVault")
+public interface VaultAccessPolicyParameters extends
+    HasInner<VaultAccessPolicyParametersInner>{
+
```
milismsft: @tiffanyachen similar comments I made about the DeletedVault interface; we should "promote" the properties from VaultAccessPolicyPropertiesInner class as methods of this interface. Also the interface should extend HasId(), HasName() and expose type() and location().

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
milismsft: @tiffanyachen If this is the case then we will have to rework a bit the "Create()" flow. What's the maximum properties in the JSON body that can be passed to the create op when CreateMode.RECOVER is specified?
Basically we will need to think of a way to shortcut the current create flow for the user to be able to use this setting. I'm thinking it might be worth having a separate set of methods as part of Vaults.java something like recoverDeletedVault(resourceGroupName, vaultName) or recoverSoftDeletedVault() (plus the Async one), instead of expecting the user to specify the CreateMode via the define()/create() flow... Because I don't see the later how it will ever succeed when we ask the user for SKU and other stuff before it can reach to withCreateMode().

anuchandy:
@tiffanyachen, @schaabs PR is good, just approved it. You can ignore the CI failures (which are not related to key vault), when we merge kv2018 back to master, we will make sure its green.

PR #444, opened 2018-05-08
ACI: refactoring and add support for launchExec() operation.

milismsft:


azure-mgmt-containerinstance/src/main/java/com/microsoft/azure/management/containerinstance/ContainerGroup.java
```diff
@@ -720,10 +746,12 @@
...
                  * Specifies the starting command lines.
                  *
-                 * @param commandLines the starting command lines the container will execute after it gets initialized
+                 * @param executable the executable which the container will call after it gets initialized
+                 * @param parameters the parameter list for the executable to be called
                  * @return the next stage of the definition
                  */
-                WithContainerInstanceAttach<ParentT> withStartingCommandLines(String... commandLines);
+                @Beta(Beta.SinceVersion.V1_11_0)
+                WithContainerInstanceAttach<ParentT> withStartingCommandLine(String executable, String... parameters);
```
yangl900: Do you want to change the other one parameter name from "commandLine" to "executable" or "pathToExecutable"? command line might be confusing.

withStartingCommandLine(String commandLine);
=>
withStartingCommandLine(String executable);

milismsft: @yangl900 Can the user pass the whole command, executable and arguments to it, as one string value? Or is the ACI service always expecting the first string to be an executable or path-to-executable?

yangl900: ACI always expect the first string to be a path to the executable. It's basically start a process in your container. You can start shell (/bin/bash), or other executable, like node.

milismsft: I made the change; thanks for the confirmation, the documentation as it comes via the Swagger spec was not very clear about that :-)

PR #447, opened 2018-05-14
NetworkACl update

tiffanyachen:
Added networkAcl updates


tiffanyachen:
Sorry - missed this one bit.

tiffanyachen:
@anuchandy @milismsft Could one of you take a quick look?

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
```diff
@@ -108,6 +108,14 @@
...
     @Beta(SinceVersion.V1_11_0)
     CreateMode createMode();
+    
+    /**
+     * Get the networkAcls value.
+     * 
+     * @return the networkAcls value
+     */
+    @Beta(SinceVersion.V1_11_0)
+    NetworkRuleSet networkAcls();
```
milismsft: @tiffanyachen In Fluent we try to avoid using acronyms especially when these are not needed. In this the method can be replaced with "networkRuleSet()". It's also consistent with a similar class found in Storage management library.

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
```diff
@@ -179,6 +187,20 @@
...
+         */
+        interface WithNetworkAcls {
+            
+            /**
+             * Set the networkAcls value.
+             *
+             * @param networkAcls the networkAcls value to set
+             * @return the next stage of key vault definition
+             */
+            WithCreate withNetworkAcls(NetworkRuleSet networkAcls);
```
milismsft: @tiffanyachen Please rename this to "withNetworkRuleSet()". Also in Java we try to avoid passing a whole complex object that needs to be built aside and instead we "decompose" them in place into a sequence of "with...()" methods corresponding to each member of the class that can be set. Think of it similar to setting the AccessPolicy.

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
```diff
@@ -298,6 +321,20 @@
...
+         */
+        interface WithNetworkAcls {
+            
+            /**
+             * Set the networkAcls value.
+             *
+             * @param networkAcls the networkAcls value to set
+             * @return the next stage of key vault definition
+             */
+            Update withNetworkAcls(NetworkRuleSet networkAcls);
```
milismsft: Same feedback as above.

azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/Vault.java
anuchandy: It seems the key vault  NetworkRuleSet is EXACTLY same as storage account network rule set. Customer will have consistent experience If we can ensure key vault nw ruleset experience is same as that storage provide. 

Storage: https://github.com/Azure/azure-libraries-for-java/blob/6f626302f38ba9da41bb00badcca66c3df97e7c8/azure-mgmt-keyvault/src/main/java/com/microsoft/azure/management/keyvault/NetworkRuleSet.java#L17

KeyVault: https://github.com/Azure/azure-libraries-for-java/blob/6f626302f38ba9da41bb00badcca66c3df97e7c8/azure-mgmt-storage/src/main/java/com/microsoft/azure/management/storage/NetworkRuleSet.java#L17 

Take a look at this “stage” in storage account definition  https://github.com/Azure/azure-libraries-for-java/blob/6f626302f38ba9da41bb00badcca66c3df97e7c8/azure-mgmt-storage/src/main/java/com/microsoft/azure/management/storage/StorageAccount.java#L446 and this “stage” in storage account update https://github.com/Azure/azure-libraries-for-java/blob/6f626302f38ba9da41bb00badcca66c3df97e7c8/azure-mgmt-storage/src/main/java/com/microsoft/azure/management/storage/StorageAccount.java#L752 

Also this helper https://github.com/Azure/azure-libraries-for-java/blob/kv2018/azure-mgmt-storage/src/main/java/com/microsoft/azure/management/storage/implementation/StorageNetworkRulesHelper.java

It seems you can copy over the helper, stages and implementations from storage. But double check the logic work for key vault as well

tiffanyachen:
Move to here https://github.com/Azure/azure-libraries-for-java/pull/449

PR #474, opened 2018-06-12
Tagging resource groups when running tests

lenala:


azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/core/ResourceGroupTaggingInterceptor.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public class ResourceGroupTaggingInterceptor implements Interceptor {
+    private static final String LOGGING_CONTEXT = "com.microsoft.azure.management.resources.ResourceGroups createOrUpdate";
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        if ("PUT".equals(chain.request().method()) && chain.request().url().uri().toString().contains("/resourcegroups/") &&
+                LOGGING_CONTEXT.equals(chain.request().header("x-ms-logging-context"))) {
+            String body = bodyToString(chain.request());
+            AzureJacksonAdapter adapter = new AzureJacksonAdapter();
```
anuchandy: should we move adaptor as a class variable instead of creating it each time we receive a request?

azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/core/ResourceGroupTaggingInterceptor.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private static String bodyToString(final Request request){
+
+        try {
+            final Request copy = request.newBuilder().build();
+            final Buffer buffer = new Buffer();
+            copy.body().writeTo(buffer);
+            return buffer.readUtf8();
+        } catch (final IOException e) {
+            return "did not work";
```
anuchandy: should we re-throw here by wrapping IOException inside runtime exception ?

azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/core/ResourceGroupTaggingInterceptor.java
```diff
@@ -0,0 +1,68 @@
...
+public class ResourceGroupTaggingInterceptor implements Interceptor {
+    private static final String LOGGING_CONTEXT = "com.microsoft.azure.management.resources.ResourceGroups createOrUpdate";
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        if ("PUT".equals(chain.request().method()) && chain.request().url().uri().toString().contains("/resourcegroups/") &&
+                LOGGING_CONTEXT.equals(chain.request().header("x-ms-logging-context"))) {
+            String body = bodyToString(chain.request());
+            AzureJacksonAdapter adapter = new AzureJacksonAdapter();
+            ResourceGroupInner rg = adapter.deserialize(body, ResourceGroupInner.class);
```
anuchandy: Should we worry about failure in adapter.deserialize method? 

if adapter.deserialize throws then we are good, we get exact reason why its failed. 

If it return null on deserialization failure then we might want to throw a runtime exception indicating deserialize failed instead of java throwing NPE in line 36.

We can test this by creating AzureJacksonAdapter instance in a test, and then try to deserialize an invalid string which is not conforming to resource group schema and see it throws vs return null.

PR #493, opened 2018-06-29
Setters and getters for network security group port ranges and address prefixes

lenala:


anuchandy:
Just summarizing our offline discussion, once we have confirmation from network team about their preference on using collection, the better solution to avoid impacting customer is:

1. Internally always use collection property instead of scalar property
2. On update when user add a new value -  
	If we find that there is already the scalar property set (previosuly) then copy it to collection and clear scalar property

This way by not using the scalar property we don't endup clearing the existing collection which can cuase sudden downtime of user application and user need to dig into logs to understand what went wrong.

PR #545, opened 2018-08-17
Update Region.java

taroyutao:
China azure has added two regions: china north2 and china east2. please add them in Region.java in next version.


jianghaolu:
Thanks for the contribution! Closing via https://github.com/Azure/azure-libraries-for-java/pull/573.

PR #562, opened 2018-08-29
[Automation] Update Maven SHA512 signature

troydai:


anuchandy:
thanks a lot @troydai 

Summarizing our discussion.

1. Each time we update maven version in the docker file, the mvn SHA needs to be updated as well, it can be retrieved from https://maven.apache.org/download.cgi
2. PR also update the script to fail fast in case of any error (by adding -e)

PR #572, opened 2018-09-10
SQL V2.0-preview

hovsepm:


api-specs.json
```diff
@@ -300,7 +300,7 @@
     "dir": "azure-mgmt-sql",
     "source": "specification/sql/resource-manager/readme.md",
     "package": "com.microsoft.azure.v2.management.sql",
-    "args": "--payload-flattening-threshold=1 --tag=package-2015-05-preview"
+    "args": "--payload-flattening-threshold=1 --tag=package-composite-v1"
```
milismsft: You should target package-composite-v3 instead since v2 is a breaking change by itself ;-)

PR #607, opened 2018-10-09
adding first set of resource samples

iscai-msft:


iscai-msft:
Done with resource samples except WebServerWithDelegatedCredentials, waiting on runtime updates

PR #611, opened 2018-10-10
Storage samples

iscai-msft:


iscai-msft:
Added all storage samples and finished resource samples by adding WebServerwithDelegatedCredentials

PR #644, opened 2018-11-07
Bumping up surefire version and setting useSystemClassLoader to false

anuchandy:
Reference: https://maven.apache.org/surefire/maven-surefire-plugin/examples/class-loading.html


anuchandy:
@jianghaolu this is an attempt to fix nightly run.

PR #701, opened 2019-02-22
Add support for UltraSSD to compute by updating compute's API to 2018-10-01

praries880:
fix for #699


azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachine.java
```diff
@@ -237,15 +239,15 @@
      * @return the refreshed instance view
      */
     @Method
-    VirtualMachineInstanceView refreshInstanceView();
+    VirtualMachineInstanceViewInner refreshInstanceView();
```
jianghaolu: We should not return Inner to the user.
If AutoRest changes this type to Inner, usually means this type is now being returned by an API directly. You will have to wrap it with an interface and add an impl.

jianghaolu: Bingo: https://github.com/Azure/azure-rest-api-specs/blob/78e1feab142c605839aa3cec23442ca26b8ef04d/specification/compute/resource-manager/Microsoft.Compute/stable/2018-10-01/compute.json#L1753

@anuchandy

praries880: done

azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachineInstanceView.java
```diff
@@ -2,335 +2,102 @@
...
-     * Set specifies the update domain of the virtual machine.
-     *
-     * @param platformUpdateDomain the platformUpdateDomain value to set
-     * @return the VirtualMachineInstanceView object itself.
-     */
-    public VirtualMachineInstanceView withPlatformUpdateDomain(Integer platformUpdateDomain) {
-        this.platformUpdateDomain = platformUpdateDomain;
-        return this;
-    }
+    Integer platformUpdateDomain();
```
jianghaolu: Don't return boxed types for interfaces. Return a default integer if null.

praries880: done

Issue #721, opened 2019-03-29
NPE when comparing VM powerstate in ExpandableStringEnum.equals()


ashbondu:
When comparing the powerstate, seeing NPE in ExpandableStringEnum.equals(). 

Syntax used is PowerState.RUNNING.equals(vm.powerState()) 
Note: The powerstate returned from Azure was null due to an intermediate error.

Stacktrace:
java.lang.NullPointerException: null
at com.microsoft.azure.management.resources.fluentcore.arm.ExpandableStringEnum.equals(ExpandableStringEnum.java:93) ~[vmw-azure-mgmt-resources-1.17.0.jar:1.17.0]

kurtzeborn:
Thank you for opening this issue! We are routing it to the appropriate team for follow up.

yaohaizh:
Fixed on PR#737.

PR #737, opened 2019-05-02
update compute to the latest API version

praries880:


praries880:
fix for  issue #721 included in this PR as well.

PR #763, opened 2019-06-20
EventHub package 2017-04 update

yaohaizh:


azure-mgmt-eventhub/src/main/java/com/microsoft/azure/management/eventhub/CaptureDescription.java
```diff
@@ -150,4 +156,24 @@ public CaptureDescription withDestination(Destination destination) {
...
+        return this.skipEmptyArchives;
+    }
+
+    /**
+     * Set a value that indicates whether to Skip Empty Archives.
+     *
+     * @param skipEmptyArchives the skipEmptyArchives value to set
+     * @return the CaptureDescription object itself.
+     */
+    public CaptureDescription withSkipEmptyArchives(Boolean skipEmptyArchives) {
```
praries880: We will need to expose this property in this file (linked to an example of how to expose it):
https://github.com/Azure/azure-libraries-for-java/blob/78e8ff2940eba34bc63f8e7be6807a377500f5c7/azure-mgmt-eventhub/src/main/java/com/microsoft/azure/management/eventhub/implementation/EventHubImpl.java#L207

Reason : The event hubs team has explicitly asked for this.

yaohaizh:
@praries880 Updated the PR

praries880:
@yaohaizh Thanks for taking care of my comments...
However i thhink i should have been more explicit, you also need to add a getter function for the configuration, for example look at this:
 https://github.com/Azure/azure-libraries-for-java/blob/06aaf5400d3570a0bd022248fc4506e1e95dcf54/azure-mgmt-eventhub/src/main/java/com/microsoft/azure/management/eventhub/EventHub.java#L56

and

https://github.com/Azure/azure-libraries-for-java/blob/06aaf5400d3570a0bd022248fc4506e1e95dcf54/azure-mgmt-eventhub/src/main/java/com/microsoft/azure/management/eventhub/implementation/EventHubImpl.java#L89

PR #770, opened 2019-07-01
 CosmosDB package 2015-04 update

yaohaizh:


yaohaizh:
Related issue: https://github.com/Azure/autorest.java/issues/375

.azure-pipelines/fluentJavaMgmt.yml
```diff
@@ -5,7 +5,7 @@ trigger:
   - master
 
 variables:
-  DefaultOptions: '--batch-mode -Dmaven.wagon.http.pool=false'
+  DefaultOptions: '-Dmaven.wagon.http.pool=false'
```
praries880: Why did we need this change?

azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/ExtendedResourceProperties.java
```diff
@@ -18,80 +18,80 @@
     /**
      * A system generated property. A unique identifier.
      */
-    @JsonProperty(value = "rid")
-    private String rid;
+    @JsonProperty(value = "_rid")
```
praries880: @anuchandy can you look at what caused this change?

azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/implementation/GremlinDatabaseInner.java
```diff
@@ -28,22 +28,22 @@
     /**
      * A system generated property. A unique identifier.
      */
-    @JsonProperty(value = "properties.rid")
-    private String rid;
+    @JsonProperty(value = "properties._rid")
```
praries880: @anuchandy same as above here... why did this naming change?

.azure-pipelines/fluentJavaMgmt.yml
yaohaizh: If with batch mode, the maven job will always run from the parent pom directory. In this case, the suppressions.xml of cosmosdb will not work.  Thus, have to exclude batch mode in this case to run check style into each individual directory.

azure-mgmt-cosmosdb/src/main/java/com/microsoft/azure/management/cosmosdb/ExtendedResourceProperties.java
yaohaizh: As I checked the cosmos-db.json, these properties are always start with `-`, not sure how the value is `rid` in our code for json property.

Which way should we honor here?

anuchandy: The swagger indeed has underscore as part of property names https://github.com/Azure/azure-rest-api-specs/blob/99d1cd23e2058369ea1d0bba1ad042f69503a18e/specification/cosmos-db/resource-manager/Microsoft.DocumentDB/stable/2015-04-08/cosmos-db.json#L3935

PR #790, opened 2019-07-25
Up runtime to 1.6.12 and okhttp to 3.14.2

jianghaolu:
Fixes https://github.com/Azure/azure-sdk-for-java/issues/3939


yaohaizh:
Should be client runtime 3.12, right?

PR #801, opened 2019-08-11
[WIP]Network: package 2019-06 update

yaohaizh:


yaohaizh:
The status is not updated, but on the pipeline it was passed. So just merged the PR.

PR #826, opened 2019-09-09
Appservice site concurrency

weidongxu-microsoft:
Change on FunctionAppImpl.java is minor rxjava style change.

Also fix sample ManageWebAppWithTrafficManager. App service plan need to be Production to support traffic manager.


yaohaizh:
What's the reason to add/remove certificates? Could we reuse the existing ones?

weidongxu-microsoft:
@yaohaizh 

Sample code use random `webapp_name` and random `domain_name`. Signed certificate is done on this particular `webapp_name.domain_name.`

Unless we remove the randomness, everytime live run would create a new certificate.

PR #835, opened 2019-09-16
Update monitor to 2019-06

weidongxu-microsoft:
No API change


azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/OdataTypeDiscriminatorTypeResolver.java
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ *
+ * Code generated by Microsoft (R) AutoRest Code Generator.
```
yaohaizh: This is code generated?

azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/Action.java
```diff
@@ -11,15 +11,17 @@
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.databind.annotation.JsonTypeResolver;
 
 /**
  * Action descriptor.
  */
-@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "odata\\.type")
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "odata.type")
```
yaohaizh: Is this manually changed or auto generated code?

azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/OdataTypeDiscriminatorTypeResolver.java
weidongxu-microsoft: Not sure. It should. Current autogenerated code regard to "odata.type" is not correct.

Autogenerated code will generate "odata\\.type", which is invalid during deserialization.
However "odata.type" will generate wrong structure (nested type inside odata) during serialization.
So the solution is to use additional OdataTypeDiscriminatorTypeResolver to deserialze as "odata.type" and serialize as "odata\\.type", which is in previous code.

Ideally autogenator should create this class.

yaohaizh: Try fix it on https://github.com/Azure/autorest.java to ramp up with autorest

weidongxu-microsoft: Actually tried locally to fix ModelTemplate.cshtml. However adding one line and another file depending on whether these is a dot in Model.BasePolymorphicDiscriminator is too much for me at that time, so pending when I got some free time...

weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-java/pull/835/commits/5203f027ae8dfc9942a933e72cac731b5dbf7bb6

API change, add multiple resource and dynamic criteria.

azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/MetricAlert.java
```diff
@@ -154,6 +176,26 @@
              * @return the next stage of metric alert definition.
              */
             WithWindowSize withTargetResource(HasId resource);
+
+            /**
+             * Sets specified resources as target to alert on metric. All resources must be of same type and in same region.
```
yaohaizh: the same type and in the same region

azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/MetricAlert.java
```diff
@@ -154,6 +176,26 @@
...
+             *
+             * @param resourceIds collection of resource id to alert on metric.
+             * @param type resource type.
+             * @param regionName resource region.
+             * @return the next stage of metric alert definition.
+             */
+            @Beta(Beta.SinceVersion.V1_25_0)
```
yaohaizh: Beta is not right. Should we update the base lib for it ?

weidongxu-microsoft: > Beta is not right. Should we update the base lib for it ?

Do you know where it is? I didn't find the annotation repository that time.

PR #836, opened 2019-09-17
Batch: regenerate code from swagger package-2019-08

xccc-msft:


azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/BatchAccountImpl.java
```diff
@@ -197,6 +192,39 @@ public void synchronizeAutoStorageKeys() {
...
+
+    @Override
+    public List<VirtualMachineFamilyCoreQuota> dedicatedCoreQuotaPerVMFamily() {
+        return this.inner().dedicatedCoreQuotaPerVMFamily();
+    }
+
+    @Override
+    public Integer dedicatedCoreQuota() {
+        PoolAllocationMode poolAllocationMode = this.inner().poolAllocationMode();
+        if(null != poolAllocationMode && !PoolAllocationMode.USER_SUBSCRIPTION.equals(poolAllocationMode.toString())){
```
weidongxu-microsoft: why use toString()?

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/BatchAccountImpl.java
```diff
@@ -7,14 +7,7 @@
...
 import com.microsoft.azure.management.apigeneration.LangDefinition;
-import com.microsoft.azure.management.batch.AccountKeyType;
-import com.microsoft.azure.management.batch.Application;
-import com.microsoft.azure.management.batch.AutoStorageBaseProperties;
-import com.microsoft.azure.management.batch.AutoStorageProperties;
-import com.microsoft.azure.management.batch.BatchAccount;
-import com.microsoft.azure.management.batch.BatchAccountCreateParameters;
-import com.microsoft.azure.management.batch.BatchAccountKeys;
-import com.microsoft.azure.management.batch.ProvisioningState;
+import com.microsoft.azure.management.batch.*;
```
weidongxu-microsoft: do not use .*. I think there is warning on it.

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/PoolImpl.java
```diff
@@ -0,0 +1,132 @@
...
+                    @Override
+                    public Pool call(PoolInner inner) {
+                        self.setInner(inner);
+                        return self;
+                    }
+                });
+    }
+
+    @Override
+    public Observable<Pool> updateResourceAsync() {
```
weidongxu-microsoft: Is it same as createResourceAsync? If so I think this could be skipped and use base class impl.

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/BatchAccount.java
```diff
@@ -81,6 +82,34 @@
...
     Map<String, Application> applications();
 
+    /**
+     * @return pools in this Batch account, indexed by name
+     */
+    Map<String, Pool> pools();
+
+    /**
+     * @return dedicatedCoreQuotaPerVMFamilyEnforced value indicating
+     * whether the core quota for the Batch Account is enforced
```
yaohaizh: `dedicatedCoreQuotaPerVMFamilyEnforced` is not necessary here

`Virtual Machine` ==> lower case `virtual machine`

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/BatchAccount.java
```diff
@@ -254,5 +299,34 @@
              */
             Update withoutApplication(String applicationId);
         }
+
+        /**
+         * The stage of a Batch account definition allowing the creation of a Batch pool.
+         */
+        interface WithPool {
+            /**
+             * Starts a definition of an pool to be created in the Batch account.
```
yaohaizh: `an` => `a`

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/BatchAccount.java
```diff
@@ -254,5 +299,34 @@
...
+             *
+             * @param poolId the reference name for the application
+             * @return the first stage of a Batch pool definition
+             */
+            Pool.UpdateDefinitionStages.Blank<Update> defineNewPool(String poolId);
```
yaohaizh: it is poolId, not sure whey the `reference name` here?

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/Pool.java
```diff
@@ -0,0 +1,162 @@
...
+             * @param mountConfigurations mount configuration value
+             * @return the next stage of the definition
+             */
+            DefinitionStages.WithConfig<ParentT> withMountConfiguration(List<MountConfiguration> mountConfigurations);
+        }
+    }
+
+    /**
+     * The entirety of a Batch pool definition as a part of parent update.
+     * @param <ParentT> the stage of the parent Batch account update to return to after attaching this definition
```
yaohaizh: this definition ==> this update

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/BatchAccountImpl.java
yaohaizh: Should not use import all to make the coding style consistently

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/PoolImpl.java
xccc-msft: We have to override here as they are declared as abstract in base class impl.

xccc-msft: We have to override here as they are declared as abstract in base class impl.

weidongxu-microsoft: See the @Override here? It is not abstract.

weidongxu-microsoft: Then just call the createResourceAsync (or let them both call private method). Point is to reduce duplicate code if possible.

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/BatchAccount.java
xccc-msft: The applicationId has comment with "the reference name for the application". I need to replace application by pool here. But for id/name, should we follow applicationId or change both this time?

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/Pool.java
```diff
@@ -30,14 +31,79 @@
...
+     */
+    List<ApplicationPackageReference> applicationPackages();
+
+    /**
+     * @return the certificates value
+     */
+    List<CertificateReference> certificates();
+
+    /**
+     * @return the size of virtual machine in the pool
```
yaohaizh: Wondering why the size is string but not scalar value?

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/BatchAccount.java
yaohaizh: Then keep the name here.

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/Pool.java
xccc-msft: The vmSize will have value like "STANDARD_D4" 
[https://github.com/Azure/azure-rest-api-specs/blob/master/specification/batch/resource-manager/Microsoft.Batch/stable/2019-08-01/examples/PoolCreate_FullExample.json#L11](url)

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/PoolsImpl.java
```diff
@@ -73,4 +74,13 @@ protected PoolImpl newChildResource(String name) {
     public void addPool(PoolImpl pool) {
         this.addChildResource(pool);
     }
+
+    public Map<String ,Pool> asMap() {
```
yaohaizh: What's the reason to provide this API?

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/PoolImpl.java
xccc-msft: Oh, it's copy-paste issue here for internal call. I will change to call updateAsync.

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/PoolsImpl.java
xccc-msft: This asMap() will be called internally by batchAccount.pools() when user wants to work on a certain pool in the map. We keep same map structure for application and pool under batch account.

azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/BatchAccount.java
xccc-msft: https://github.com/Azure/azure-rest-api-specs/blob/master/specification/batch/resource-manager/Microsoft.Batch/stable/2019-08-01/BatchManagement.json#L2034
should we change to lower case?

Issue #809, opened 2019-08-20
[BUG] CI pipeline is not stable


yaohaizh:
Need fix the pipeline issue for Java regarding surefire in ADO

yaohaizh:
Close this one as it is maven central not stable.

PR #844, opened 2019-09-25
maven mirror

weidongxu-microsoft:


weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-java/issues/809

Issue #823, opened 2019-09-05
[FEATURE REQ] Add enableCassandraConnector & connectorOffer for CosmosDB


yaohaizh:
https://github.com/Azure/azure-rest-api-specs/pull/6550

yaohaizh:
Postpone as the service is not totally ready yet.

yaohaizh:
Merged.

PR #845, opened 2019-09-25
Add Cassandra Connector to cosmos db account

ChenTanyi:
#823 
* Cannot pass test, all subscription return error:
```
com.microsoft.azure.CloudException: EnableCassandraConnector is not supported for the target subscription
```


ChenTanyi:
service have supported

PR #849, opened 2019-09-26
Fix traffic analytics configuration properties

weidongxu-microsoft:


weidongxu-microsoft:
The change on FlowLogSettingsImpl.java is to mitigate an existing service issue of sending empty flowAnalyticsConfiguration.

azure-mgmt-network/src/main/java/com/microsoft/azure/management/network/TrafficAnalyticsConfigurationProperties.java
```diff
@@ -23,19 +23,19 @@
     /**
      * The resource guid of the attached workspace.
      */
-    @JsonProperty(value = "workspaceId", required = true)
+    @JsonProperty(value = "workspaceId")
```
yaohaizh: Please document this changes.

weidongxu-microsoft: Swagger network 2019-07 would have this fix.

PR #850, opened 2019-09-26
Proxy update

xccc-msft:


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ProviderRegistrationInterceptor.java
```diff
@@ -40,6 +42,18 @@ public ProviderRegistrationInterceptor(AzureTokenCredentials credentials) {
...
+
+    public Response interceptWithProxy(Chain chain, Proxy proxy) throws IOException {
+        return intercept(chain, proxy, null);
+    }
+
+    public Response interceptWithProxyAuthenticator(Chain chain, Authenticator authenticator) throws IOException {
+        return intercept(chain, null, authenticator);
+    }
+
+    private Response intercept(Chain chain, Proxy proxy, Authenticator authenticator) throws IOException {
```
weidongxu-microsoft: I assume `intercep(Chain chain)` is called by okhttp, so I am not sure whether the new method would be called or not.

PR #859, opened 2019-10-09
Appgwv2

ovidiu10:


yaohaizh:
@xccc-msft  please work with @ovidiu10 for the new fluent sample.

Issue #847, opened 2019-09-25
[BUG] Generate random password & SSHKey for ManageNetworkInterface.java & ManageNetworkSecurityGroup


yaohaizh:
Replace the password/SSHKey with random generate values.

PR #863, opened 2019-10-17
fix: randomize password and sshkey

ChenTanyi:
#847


yaohaizh:
@ChenTanyi  Need fix test case. Might need intercept for the validation of password/SSHKey in the response.

PR #883, opened 2019-10-31
Resources: support WhatIf deployment

xccc-msft:


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/WhatIfOperationResultImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public String status() {
+        return this.inner().status();
+    }
+
+    @Override
+    public List<WhatIfChange> changes() {
+        return this.inner().changes();
```
weidongxu-microsoft: use Collections.unmodifiableList if possible.

Issue #881, opened 2019-10-31
[FEATURE REQ] Function App support Linux Java 8


weidongxu-microsoft:
Function App support Linux Java 8

```
{
  "type": "Microsoft.Web/sites",
  "kind": "functionapp,linux",
  "location": "West US",
  "properties": {
    "reserved": true,
    "siteConfig": {
      "linuxFxVersion":"java|8",
      "appSettings": [
          {
              "name": "FUNCTIONS_EXTENSION_VERSION",
              "value": "~2"
          },
          {
             "name": "AzureWebJobsStorage",
             "value": "DefaultEndpointsProtocol=https;AccountName=functionstestinb489;AccountKey=XXX;EndpointSuffix=core.windows.net"
          },
          {
            "name": "WEBSITE_USE_ZIP",
            "value": "https://functionstests.blob.core.windows.net/public/docker/java-functions.zip"
          },
          {
            "name": "FUNCTIONS_WORKER_RUNTIME",
            "value": "java"
          }
      ]
    }
  }
}
```

```
{
  "type": "Microsoft.Web/sites",
  "kind": "functionapp,linux",
  "location": "North Europe",
  "properties": {
    "reserved": true,
    "serverFarmId": "/subscriptions/{}/resourcegroups/linuxarm/providers/Microsoft.Web/serverfarms/test-north-europ",
    "alwaysOn": true,
    "siteConfig": {
      "linuxFxVersion":"DOCKER|mcr.microsoft.com/azure-functions/java:2.0-java8-appservice",
      "appSettings": [
          {
              "name": "FUNCTIONS_EXTENSION_VERSION",
              "value": "~2"
          },
          {
             "name": "AzureWebJobsStorage",
             "value": "DefaultEndpointsProtocol=https;AccountName=storageaccountnodes93b7;AccountKey=;EndpointSuffix=core.windows.net"
          },
          {
            "name": "FUNCTIONS_WORKER_RUNTIME",
            "value": "java"
          }
      ]
    }
  }
}
```

weidongxu-microsoft:
Fixed in 1.28. To be released on Nov. 15.

PR #889, opened 2019-11-06
Support function app linux java8

weidongxu-microsoft:


weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-java/issues/881

weidongxu-microsoft:
@Flanker32 Please take a look on this PR for funciton app linux 8. And let us know if the interface is OK for you.

There is 3 test cases.

https://github.com/Azure/azure-libraries-for-java/blob/ee9af8ba6c5a7eb15c8f7e9181dbd3bee3d3e321/azure-mgmt-appservice/src/test/java/com/microsoft/azure/management/appservice/FunctionAppsTests.java#L165
This one is to create 1 function app with new consumption plan, 1 function app with new app service plan, 1 function app sharing the app service plan.

https://github.com/Azure/azure-libraries-for-java/blob/ee9af8ba6c5a7eb15c8f7e9181dbd3bee3d3e321/azure-mgmt-appservice/src/test/java/com/microsoft/azure/management/appservice/FunctionAppsTests.java#L234
This one is to create 1 function app with new elastic premium plan.

https://github.com/Azure/azure-libraries-for-java/blob/ee9af8ba6c5a7eb15c8f7e9181dbd3bee3d3e321/azure-mgmt-appservice/src/test/java/com/microsoft/azure/management/appservice/FunctionAppsTests.java#L261
This one is to create 1 function app with new elastic premium plan, using private docker container. It would require pre-setup of container registry (e.g. weidxuregistry.azurecr.io) and pre-populated docker image (e.g. weidxuregistry.azurecr.io/az-func-java:v1).

weidongxu-microsoft:
Zip deploy not working in playback. I will update the test cases.

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/FunctionApp.java
```diff
@@ -136,11 +138,18 @@
          */
         interface Blank extends DefinitionWithRegion<NewAppServicePlanWithGroup> {
             /**
-             * Uses an existing app service plan for the function app.
+             * Uses an existing Windows app service plan for the function app.
              * @param appServicePlan the existing app service plan
              * @return the next stage of the definition
              */
             ExistingAppServicePlanWithGroup withExistingAppServicePlan(AppServicePlan appServicePlan);
```
yaohaizh: The naming should be consistent.

weidongxu-microsoft: You mean function naming or docstring naming?

yaohaizh: function naming. Second though, we need back compatible with previous API, so that the change might not necessary. But anyway, user can be confused in the case.

weidongxu-microsoft:
1.28.0 released.

PR #895, opened 2019-11-13
Resources: fix deleteById in policy definition operations

xccc-msft:


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/PolicyDefinitionsImpl.java
```diff
@@ -44,6 +44,10 @@ public PolicyDefinitionImpl getByName(String name) {
 
     @Override
     public Completable deleteByIdAsync(String name) {
+        String[] parts = name.split("/");
```
yaohaizh: Curious why need the last part of the name? Will the name have conflict?

Issue #903, opened 2019-11-19
[FEATURE REQ] Update compute API version to 2019-07-01


yaohaizh:
Here is the link to the swagger changes: https://github.com/Azure/azure-rest-api-specs/pull/7748

Below are the new changes required for the SDK:
1.	Add 'Spot' enum value to 'Priority' enum.
2.	Can you confirm we are allowing the update of ‘Priority’ and ‘MaxPrice’ for both VM and VMSS? Some context below
•	We now support update of priority only between 'Low' and 'Spot' for both VM and VMSS. Note: Update of priority between 'Low' and 'Regular', also between 'Spot' and 'Regular' are not allowed.
•	VM and VMSS support the update of MaxPrice only when the VM or VMSS are in deallocated state.

yaohaizh:
This has been fixed.

PR #908, opened 2019-11-27
Compute: update 2019 07 01

ChenTanyi:
#903 
@yaohaizh here is the test:
for VM, maxPrice cannot be updated when VM is in deallocated state, priority can be update between LOW and SPOT, but cannot update between REGULAR and others.
for VMSS, maxPrice can be updated any time, but the priority cannot be updated, since swagger doesn't have such field. https://github.com/Azure/azure-rest-api-specs/blob/f57417493691c0d9a15440ffd7fdb4173bf7295c/specification/compute/resource-manager/Microsoft.Compute/stable/2019-07-01/compute.json#L9200


azure-mgmt-compute/src/test/java/com/microsoft/azure/management/compute/VirtualMachineOperationsTests.java
```diff
@@ -207,12 +207,49 @@ public void canCreateLowPriorityVirtualMachine() throws Exception {
...
+        try {
+            foundVM.update()
+                    .withMaxPrice(1500.0)
+                    .apply();
+            // not run to assert
+            Assert.assertEquals((Double) 1500.0, foundVM.billingProfile().maxPrice());
+            Assert.assertTrue(false);
+        } catch (AssertionError e) {
+            throw e;
+        } catch (Exception e) {} // cannot change max price when vm is running
```
weidongxu-microsoft: Maybe better to catch CloudException (and then there would be no need to catch-rethrow AssertionError)? No need to record again.

azure-mgmt-compute/src/test/java/com/microsoft/azure/management/compute/VirtualMachineOperationsTests.java
```diff
@@ -207,12 +207,49 @@ public void canCreateLowPriorityVirtualMachine() throws Exception {
...
+        try {
+            foundVM.update()
+                    .withPriority(VirtualMachinePriorityTypes.REGULAR)
+                    .apply();
+            // not run to assert
+            Assert.assertEquals(VirtualMachinePriorityTypes.REGULAR, foundVM.priority());
+            Assert.assertTrue(false);
+        } catch (AssertionError e) {
+            throw e;
+        } catch (Exception e) {} // cannot change priority from spot to regular
```
weidongxu-microsoft: Just asking. Is there any service message saying that cannot be done and provide reason?

ChenTanyi: No, the exception is just: `com.microsoft.azure.CloudException: Changing property 'priority' is not allowed.` These contexts are all from services email.

ChenTanyi:
disk import data error due to service 400, cannot be merged

azure-mgmt-compute/src/test/java/com/microsoft/azure/management/compute/VirtualMachineOperationsTests.java
```diff
@@ -244,7 +244,7 @@ public void canCreateUpdatePriorityAndPrice() throws Exception {
             // not run to assert
             Assert.assertEquals(VirtualMachinePriorityTypes.REGULAR, foundVM.priority());
             Assert.fail();
-        } catch (CloudException e) {} // cannot change priority from spot to regular
+        } catch (CloudException e) {} // cannot change priority from low to regular
```
yaohaizh: Please confirm this behavior with compute service team

ChenTanyi: This is written by service team, and my test also fits their statement.

PR #914, opened 2019-12-10
update linux java8 function image to latest v3

weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-java/issues/918


weidongxu-microsoft:
Function app service will support v2 at least until next Jan. On Jan. v3 would be the default.

Issue #875, opened 2019-10-25
[FEATURE REQ] update appservice to package-2019-08


weidongxu-microsoft:
update appservice to package-2019-08

weidongxu-microsoft:
Blocked by service team.

PR #933, opened 2019-12-24
appservice update

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/875

Add a test case.
https://github.com/Azure/azure-libraries-for-java/commit/48c820906b3139e968eb9fa8de77ba8fd83ff65c

Add new enum, older ones are deprecated.
https://github.com/Azure/azure-libraries-for-java/commit/eb3ddb507edfedf5e2ae36102b9b56986c24c20f

Add FileProvider for different behavior when generate certificate/pfx in live/record/playback.
https://github.com/Azure/azure-libraries-for-java/commit/dccab699ab8fe4bd921ae066cf341e30ebb47fa9


azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/CertificatePatchResource.java
```diff
@@ -376,4 +382,24 @@ public CertificatePatchResource withServerFarmId(String serverFarmId) {
         return this;
     }
 
+    /**
+     * Get cNAME of the certificate to be issued via free certificate.
```
yaohaizh: cName or CNAME? be consistent

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/CloningInfo.java
```diff
@@ -55,6 +55,12 @@
     @JsonProperty(value = "sourceWebAppId", required = true)
     private String sourceWebAppId;
 
+    /**
+     * Location of source app ex: West US or North Europe.
```
yaohaizh: Does this value mapping to the Region? If so, use region instead to avoid wrong typing of location

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/CertificatePatchResource.java
weidongxu-microsoft: Should be CNAME.

Seems autorest.java issue. Swagger is good.

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/CloningInfo.java
weidongxu-microsoft: This class is auto-generated. And we do not have wrapper class for it.

Issue #934, opened 2019-12-25
[BUG] input param of withPrivateRegistryImage is ambiguous


weidongxu-microsoft:
**Describe the bug**
`WithCredentials withPrivateRegistryImage(String imageAndTag, String serverUrl);`

Hard to tell whether `imageAndTag` would include host or not.

***Exception or Stack Trace***
Add the exception log and stack trace if available

**To Reproduce**

***Code Snippet***
Add the code snippet that causes the issue.

**Expected behavior**

**Screenshots**

**Setup (please complete the following information):**
 - IDE : [e.g. IntelliJ]
 - Version of the Library used

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ ] Bug Description Added
- [ ] Repro Steps Added
- [ ] Setup information Added

Issue #915, opened 2019-12-10
[BUG] SDK will override users' app settings when create new functions


Flanker32:
**Describe the bug**
SDK will override users' app settings(`AzureWebJobsStorage`,`AzureWebJobsDashboard`) when create new functions as SDK will create storage account asynchronous and overwrite the appsetting once storage account created

**To Reproduce**

Create function with fluent SDK and define `AzureWebJobsStorage` or `AzureWebJobsDashboard` before execute `create()`

***Code Snippet***
``` java
    protected void createFunctionApp() throws AzureAuthFailureException, MojoExecutionException 
    {
        ...
        final FunctionRuntimeHandler runtimeHandler = getFunctionRuntimeHandler();
        final WithCreate withCreate = runtimeHandler.defineAppWithRuntime();
        configureAppSettings(withCreate::withAppSettings, getAppSettingsWithDefaultValue());
        withCreate.withJavaVersion(DEFAULT_JAVA_VERSION).withWebContainer(null).create();
        ...
    }

    protected void configureAppSettings(final Consumer<Map> withAppSettings, final Map appSettings) {
        if (appSettings != null && !appSettings.isEmpty()) {
            withAppSettings.accept(appSettings);
        }
    }
```

**Expected behavior**
Use `AzureWebJobsStorage` and `AzureWebJobsDashboard` specified by users

**Setup (please complete the following information):**
 - OS:  Windows 10
 - IDE : IntelliJ 2019.2
 - Version of the Library used : 1.28.0

**Additional context**
SDK could check whether those app settings exists before deploy and update it once user didn't set it.

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Bug Description Added
- [x] Repro Steps Added
- [x] Setup information Added

yaohaizh:
@weidongxu-microsoft could you help take a look at this issue?

PR #936, opened 2019-12-25
fix webapp (1) do not override appsetting if provided by user; (2) try add docker registry server to image name

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/915
https://github.com/Azure/azure-libraries-for-java/commit/17bccb185cca100db6b0bf05a731d3430ec9e2c8

Fix https://github.com/Azure/azure-libraries-for-java/issues/934
https://github.com/Azure/azure-libraries-for-java/commit/f7b604032d982ea3d586809f1267262b1cf9413c


weidongxu-microsoft:
Service will actually verify WEBSITE_CONTENTAZUREFILECONNECTIONSTRING

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/AppServiceBaseImpl.java
```diff
@@ -436,6 +439,73 @@ public FluentImplT withExistingAppServicePlan(AppServicePlan appServicePlan) {
         return withOperatingSystem(appServicePlanOperatingSystem(appServicePlan));
     }
 
+    public FluentImplT withPublicDockerHubImage(String imageAndTag) {
+        ensureLinuxPlan();
```
Flanker32: As users could create docker app service with windows runtime in portal, i think we should not check the app service plan here.

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/AppServiceBaseImpl.java
```diff
@@ -436,6 +439,73 @@ public FluentImplT withExistingAppServicePlan(AppServicePlan appServicePlan) {
...
+
+    private static String smartCompletionPrivateRegistryImage(String imageAndTag, String serverUrl) {
+        try {
+            URL url = new URL(serverUrl);
+            String registryServer = url.getAuthority();
+            if (!registryServer.isEmpty() && !imageAndTag.trim().startsWith(registryServer)) {
+                String[] segments = imageAndTag.split(Pattern.quote("/"));
+                if (segments.length == 1) {
+                    // it appears that imageAndTag does not contain registry server, add registry server before it.
+                    imageAndTag = String.format("%s/%s", registryServer, imageAndTag);
```
ChenTanyi: I think it could add url path behind `registryServer` rather than only `/`

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/AppServiceBaseImpl.java
```diff
@@ -436,6 +439,73 @@ public FluentImplT withExistingAppServicePlan(AppServicePlan appServicePlan) {
...
+                }
+                if (segments.length > 1) {
+                    String segment = segments[0];
+                    if (!segment.isEmpty() && !segment.contains(".") && !segment.contains(":") && !segment.equals(registryServer)) {
```
ChenTanyi: `.` can be in image name, and `:` can not be dealt by docker in docker/kitematic #1840

ChenTanyi: This is just a tool for docker, can be ignored.

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/AppServiceBaseImpl.java
weidongxu-microsoft: I see. Web app got it. But function app still limit to Linux so far. I will no longer check this for web app.

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/AppServiceBaseImpl.java
weidongxu-microsoft: OK. path added.

Issue #898, opened 2019-11-15
[FEATURE REQ] sample for function app linux java 8 end2end flow


weidongxu-microsoft:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ ] Description Added
- [ ] Expected solution specified

weidongxu-microsoft:
Delay to next milestone.

Service has yet to finalize java v3 image. They do not want user to use v2 image, which currently no one uses, and service wishes to get deprecated very soon.

Maven plugin team would make a customized function runtime definition which would enable them to use v3 preview. However our typical user probably will not do that.

So I figure milestone 162 is not the best time to add sample code to encourage user to use this.

PR #941, opened 2019-12-30
Linux function app sample

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/898

Sample for Linux function app, use recommended "run from a package" approach.
https://docs.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package


yaohaizh:
Need add this to smaple.json file for publishing.

PR #942, opened 2019-12-31
Fix test cases

yaohaizh:


azure-mgmt-resources/src/main/java/com/azure/management/RestClient.java
```diff
@@ -8,31 +11,40 @@
...
+    private final URL baseUrl;
+
+    /**
+     * The original builder for this rest client.
+     */
+    private final RestClientBuilder builder;
 
-    public RestClient(RestAsyncClient client) {
-        this.client = client;
+    public RestClient(URL baseUrl, HttpPipeline pipeline, RestClientBuilder builder) {
```
weidongxu-microsoft: Not usual to take a Builder as parameter for constructor. And if we go from Builder, we probably do not need a public constructor for RestClient.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestUtilities.java
```diff
@@ -24,4 +27,12 @@ public static void sleep(int milliseconds, boolean isRecordMode) {
         }
     }
 
+    public static <T> int getPagedIterableSie(PagedIterable<T> iterable) {
```
weidongxu-microsoft: Size ?

azure-mgmt-resources/src/test/java/com/azure/management/resources/fluentcore/dag/ProxyTaskGroupTests.java
```diff
@@ -1451,10 +1452,14 @@ public void beforeGroupInvoke() {
...
         final ArrayList<String> seen = new ArrayList<>();
+        CountDownLatch down = new CountDownLatch(1);
         itiC.getTaskGroup()
                 .invokeAsync(itiC.getTaskGroup().newInvocationContext())
                 .subscribe(indexable ->
-                        seen.add(indexable.getKey()));
+                                seen.add(indexable.getKey()),
+                        throwable -> down.countDown(),
+                        () -> down.countDown());
+        down.await();
```
weidongxu-microsoft: Should we use await with timeout?

yaohaizh: Without network connection, just keep the case simple.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestUtilities.java
yaohaizh: Fixed

azure-mgmt-resources/src/main/java/com/azure/management/RestClient.java
yaohaizh: Agree, this constructor is migrated all the V1.  We need sort out of the build/serviceclient relationship.

PR #944, opened 2020-01-02
vnext update gulp

weidongxu-microsoft:
Temporary put dirty handling of "tag" into gulp.json.

It creates a temporary md file which contains only the input files from the specified "tag".

Usage
`gulp codegen --spec-root "C:\github\azure-rest-api-specs" --autorest-java "C:\github_fork\autorest.java\fluentgen" --projects resources`


gulpfile.js
```diff
@@ -67,7 +67,7 @@ var autoRestExe;
 gulp.task('codegen', function(cb) {
     if (autoRestVersion.match(/[0-9]+\.[0-9]+\.[0-9]+.*/) ||
         autoRestVersion == 'latest') {
-            autoRestExe = 'autorest ---version=' + autoRestVersion;
+            autoRestExe = 'autorest-beta ---version=' + autoRestVersion;
```
yaohaizh: Need update package.json

gulpfile.js
```diff
@@ -107,8 +107,13 @@ var codegen = function(project, cb) {
         deleteFolderRecursive(sourcesToDelete);
     }
 
-    console.log('Generating "' + project + '" from spec file ' + specRoot + '/' + mappings[project].source);
-    var generator = '--fluent';
+    let readmeFile = specRoot + '/' + mappings[project].source;
```
yaohaizh: use path.join for multi platform

gulpfile.js
weidongxu-microsoft: I think package.json is not needed. It just calls autorest-beta on exec. No dependency on nodejs.

gulpfile.js
weidongxu-microsoft: There were a comment nearby: "path.join won't work if specRoot is a URL", so it uses '/'.

For now the dirty approach would need the spec on local filesystem to work. But still keep it this way. Some day we might no longer need this hacking and re-enable the URL.

gulpfile.js
```diff
@@ -144,6 +150,42 @@ var codegen = function(project, cb) {
     return execa(cmd, [], { shell: true, stdio: "inherit" });
 };
 
+var findTag = function(stringContainsTag) {
+    let regex = new RegExp('--tag=(\\S+)');
```
yaohaizh: const

PR #946, opened 2020-01-03
Add storage & keyvault project.

yaohaizh:


yaohaizh:
> Is it "keyvault" or "arm-keyvault" in api-specs.json?

Should be  arm-keyvault. The previous is not generated from the current implementation.

Issue #953, opened 2020-01-10
[BUG] Get app service by resource group will throw exception if not exists


Flanker32:
**Describe the bug**
Will get exception when execute `WebApps.getByResourceGroup(String resourceGroup, String appName)` for an nonexistent webapp instead of null;

***Exception or Stack Trace***
```
Exception in thread "main" java.util.NoSuchElementException: Sequence contains no elements
	at rx.internal.operators.OperatorSingle$ParentSubscriber.onCompleted(OperatorSingle.java:115)
	at rx.internal.operators.DeferredScalarSubscriber.complete(DeferredScalarSubscriber.java:83)
	at rx.internal.operators.OnSubscribeTakeLastOne$TakeLastOneSubscriber.onCompleted(OnSubscribeTakeLastOne.java:55)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.emitLoop(OperatorMerge.java:656)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.emit(OperatorMerge.java:568)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.onCompleted(OperatorMerge.java:281)
	at rx.internal.operators.OnSubscribeMap$MapSubscriber.onCompleted(OnSubscribeMap.java:97)
	at rx.internal.operators.OnSubscribeMap$MapSubscriber.onCompleted(OnSubscribeMap.java:97)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.emitLoop(OperatorMerge.java:656)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.emit(OperatorMerge.java:568)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.onCompleted(OperatorMerge.java:281)
	at rx.internal.operators.OnSubscribeMap$MapSubscriber.onCompleted(OnSubscribeMap.java:97)
	at retrofit2.adapter.rxjava.CallArbiter.deliverResponse(CallArbiter.java:144)
	at retrofit2.adapter.rxjava.CallArbiter.emitResponse(CallArbiter.java:102)
	at retrofit2.adapter.rxjava.CallExecuteOnSubscribe.call(CallExecuteOnSubscribe.java:46)
	at retrofit2.adapter.rxjava.CallExecuteOnSubscribe.call(CallExecuteOnSubscribe.java:24)
	at rx.Observable.unsafeSubscribe(Observable.java:10327)
[...]

PR #957, opened 2020-01-13
fix NoSuchElementException in WebApp and FunctionApp

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/953

`Observable.just(null)` probably is not a good idea. But anyway `null` is what returned from `Inner.getByResourceGroupAsync()`.

### DIFF
--- a/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/WebAppsImpl.java
+++ b/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/WebAppsImpl.java
@@ -63,7 +63,7 @@ class WebAppsImpl
             @Override
             public Observable<WebApp> call(final SiteInner siteInner) {
                 if (siteInner == null) {
-                    return null;
+                    return Observable.just(null);
                 }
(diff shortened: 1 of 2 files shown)


azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/WebAppsImpl.java
```diff
@@ -63,7 +63,7 @@ public WebApp call(SiteConfigResourceInner siteConfigResourceInner, SiteLogsConf
             @Override
             public Observable<WebApp> call(final SiteInner siteInner) {
                 if (siteInner == null) {
-                    return null;
```
yaohaizh: Should be Observable.complete()

weidongxu-microsoft: Seems no such thing as `Observable.complete()`?

`null` is actually OK for rxjava 1.x, which is used here. rxjava 2.x would reject null.

PR #963, opened 2020-01-15
Add record and playback test

ChenTanyi:


azure-mgmt-resources/src/main/java/com/azure/management/RestClientBuilder.java
```diff
@@ -144,6 +148,11 @@ public RestClientBuilder withConfiguration(Configuration configuration) {
         return this;
     }
 
+    public RestClientBuilder withClientId(String clientId) {
```
weidongxu-microsoft: Need a bit more discussion on this interface, whether need to group into withCredential?

azure-mgmt-resources/src/main/java/com/azure/management/RestClientBuilder.java
```diff
@@ -66,7 +69,8 @@ public RestClient buildClient() {
         // Closest to API goes first, closest to wire goes last.
         final List<HttpPipelinePolicy> policies = new ArrayList<>();
         // TODO: Add UserAgent
-        // TODO Add credential policy
+        // policies.add(new UserAgentPolicy("test", this.getClass().getPackage().getName(), this.getClass().getPackage().getImplementationVersion(), Configuration.getGlobalConfiguration()));
+        policies.add(new BearerTokenAuthenticationPolicy(this.credential, (this.clientId == null ? "" : this.clientId) + "/.default"));
```
yaohaizh: The second part should be factor out and make it clear of the meaning.

ChenTanyi:
@weidongxu-microsoft @yaohaizh The record and playback test can pass the LRO.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/InterceptorManager.java
```diff
@@ -37,7 +36,8 @@
 public class InterceptorManager {
 
     private final static String RECORD_FOLDER = "session-records/";
-    private static final String BODY_LOGGING = "x-ms-body-logging";
+    private final static String BODY_LOGGING = "x-ms-body-logging";
+    private final static Integer HTTP_TEMPORARY_REDIRECT = 307; // HTTP Status Code
```
weidongxu-microsoft: just int?

azure-mgmt-resources/src/main/java/com/azure/management/BearerTokenAuthenticationPolicy.java
```diff
@@ -72,7 +72,8 @@ private String getDefaultScopeFromRequest(HttpRequest request) {
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
-        if ("http".equals(context.getHttpRequest().getUrl().getProtocol().toLowerCase())) {
+        if ("http".equals(context.getHttpRequest().getUrl().getProtocol().toLowerCase()) &&
+            !"localhost".equals(context.getHttpRequest().getUrl().getHost().toLowerCase().split(":")[0])) {
```
weidongxu-microsoft: Try see if we can have another solution? It seems not reasonable to change product code because of test reason.

ChenTanyi: I use it due to I thought the `localhost` should not be blocked even there is user use it in real request. And the request for `localhost` usually doesn't have `https`.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/InterceptorManager.java
ChenTanyi: The original one is int https://hc.apache.org/httpcomponents-core-4.4.x/httpcore/apidocs/org/apache/http/HttpStatus.html

weidongxu-microsoft: Got it. I mean just use `int` should be OK, no need for `Integer` (which could have problem when using ==).

ChenTanyi: Got it, I have seen the `Integer.ParseInt`, so I thought it would get Integer before.

azure-mgmt-resources/src/main/java/com/azure/management/BearerTokenAuthenticationPolicy.java
weidongxu-microsoft: Yes, probably OK. Just check whether we could say use an overloaded BearerTokenPolicy for playback (or do we need Auth on playback at all?).

ChenTanyi: I think we could add a `skipAuthetication` for `restClientBuilder`.

ChenTanyi:
@weidongxu-microsoft @yaohaizh I use the credential to check if need to skip authentication or not. Is it OK? Do we need to a new function like `withoutCredential` to force skip authentication?

weidongxu-microsoft:
@ChenTanyi 
I think it is OK (better not to interfere with Token since I am not good on security).

PR #964, opened 2020-01-15
fix constant in model property, but might not complete; fix client faltten

weidongxu-microsoft:
Related issue in autorest.java https://github.com/Azure/autorest.java/issues/483

This might not correct. I've added the constant regardless of whether the property is required or not.


yaohaizh:
Done

PR #967, opened 2020-01-16
Add User Agent Policy in Management SDK

ChenTanyi:


azure-mgmt-resources/src/main/java/com/azure/management/ManagementUserAgentPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+        if (sdkVersion.isEmpty()) {
+            sdkVersion = defaultSdkVersion;
+        }
+
+        String applicationId = context.getData(APPLICATION_ID_KEY).orElse("").toString();
+        if (applicationId.isEmpty()) {
+            applicationId = httpLogOptions.getApplicationId();
+        }
+
+        return new UserAgentPolicy(applicationId, sdkName, sdkVersion, configuration).process(context, next);
```
weidongxu-microsoft: Maybe another way to reuse this part? "new" everytime seems not a good practise.

ChenTanyi: The functions in this policy are private, maybe I copy it if we don't use `new`.

PR #973, opened 2020-01-20
feat: remove dependency for org.apache.commons and add azure-core-http-netty 1.2.0

ChenTanyi:


azure-mgmt-resources/src/main/java/com/azure/management/UserAgentPolicy.java
```diff
@@ -86,31 +85,35 @@ private boolean telemetryDisabled() {
         return configuration.get(Configuration.PROPERTY_AZURE_TELEMETRY_DISABLED, false);
     }
 
+    private boolean isEmptyString(String s) {
```
weidongxu-microsoft: static ?

ChenTanyi: done

PR #974, opened 2020-01-20
GraphRbac: update V2 Graph with azure core

xccc-msft:


azure-mgmt-graph-rbac/pom.xml
```diff
@@ -47,39 +47,45 @@
...
+      <artifactId>azure-identity</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.microsoft.azure</groupId>
-      <artifactId>azure-client-authentication</artifactId>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-mgmt-resources</artifactId>
+      <version>2.0.0-SNAPSHOT</version>
+    </dependency>
+    <dependency>
```
yaohaizh: If the dependency is common to all the project, please factor out the the parent projects in the dependencymanagement section.

PR #976, opened 2020-01-20
Add authentication policy

ChenTanyi:


azure-mgmt-resources/src/main/java/com/azure/management/RestClientBuilder.java
```diff
@@ -39,6 +40,9 @@
     private HttpLogOptions httpLogOptions;
     private Configuration configuration;
     private SerializerAdapter serializerAdapter;
+    private List<String> scopes;
+
+    private static final String[] defaultScopes = new String[] { AzureEnvironment.AZURE.getResourceManagerEndpoint() + "/.default" };
```
weidongxu-microsoft: Add TODO, should use endpoint from AzureEnvironment from credential.

azure-mgmt-resources/src/main/java/com/azure/management/AzureTokenCredential.java
```diff
@@ -43,33 +43,33 @@ public AzureTokenCredential(AzureEnvironment environment, String domain) {
      * @param request the details of the token request
      * @return a Publisher that emits a single access token
      */
-    @Override
-    public final Mono<AccessToken> getToken(TokenRequestContext request) {
```
weidongxu-microsoft: OK for now, please investigate how other implementation of this class do on this `getToken`

PR #987, opened 2020-01-23
vnext keyvault fix

weidongxu-microsoft:


weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-java/pull/987/commits/2d064bb5cfef28d058161248c1acae3dfbe55074 is PagedFlux to PagedFlux with inner flatMap, which depends on core 1.3.0-beta.1. So code is commented out.

azure-mgmt-keyvault/src/main/java/com/azure/management/keyvault/implementation/Utils.java
```diff
@@ -0,0 +1,26 @@
+package com.azure.management.keyvault.implementation;
```
yaohaizh: Header comment

PR #998, opened 2020-02-03
Convert Network v2

yaohaizh:


yaohaizh:
I have completed the conversion of network since it is required by other modules. Still need fix tests.

PR #999, opened 2020-02-03
add delegate in base to handle get404 as null

weidongxu-microsoft:
In v1, GET method with 404 response is treated as null

Whether we need to do this in v2? Currently, 404 is exception.


yaohaizh:
We should surface the 404 to the customer and let the customer handle it. Thanks.

weidongxu-microsoft:
OK.

This would be a huge backward incompatible. Basically all getXXX would fall on it.

PR #1002, opened 2020-02-04
feat: rewrite authentication policy to use default scope

ChenTanyi:


weidongxu-microsoft:
If possible could you link the related class in v1? Easier to see the difference.

Also does it main I no longer need to config scope for RestClient on data-plane? And does it mean management-plane and data-plane and share the same RestClient?

azure-mgmt-resources/src/main/java/com/azure/management/BearerTokenAuthenticationPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+                    break;
+                }
+            }
+        }
+        return resource + "/.default";
+    }
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        if ("http".equals(context.getHttpRequest().getUrl().getProtocol())) {
```
weidongxu-microsoft: Is it possible to have UPPERCASE of getProtocol?

ChenTanyi:
The base class is https://github.com/azure/azure-sdk-for-java/blob/6a10de8a0afbc50532ff084c04d41737e3785dbc/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/BearerTokenAuthenticationPolicy.java#L12
The get default scope is copy from AzureTokenCredential https://github.com/Azure/azure-libraries-for-java/blob/6850012dd01301c375050d54632527693d08eef0/azure-mgmt-resources/src/main/java/com/azure/management/AzureTokenCredential.java#L47-L72
I have tested with Graph.RBAC, we don't need to add default scope for data-plane any more.

azure-mgmt-resources/src/main/java/com/azure/management/BearerTokenAuthenticationPolicy.java
ChenTanyi: I haven't met a UPPERCASE protocol url, but I think we could handle it.

azure-mgmt-resources/src/main/java/com/azure/management/BearerTokenAuthenticationPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+            scopes = new String[] {getDefaultScopeFromRequest(context.getHttpRequest())};
+        } else {
+            scopes = this.scopes;
+        }
+        assert scopes.length > 0;
+
+        Mono<AccessToken> tokenResult;
+        AccessToken token = tokenCache.get(scopes[0]);
+        if (token == null || token.isExpired()) {
+            tokenResult = this.credential.getToken(new TokenRequestContext().addScopes(scopes)).doOnNext(accessToken -> this.tokenCache.put(scopes[0], accessToken));
```
weidongxu-microsoft: Not sure about `doOnNext`, it could still follow by an Error instead of Complete. Maybe `doOnSuccess`?

ChenTanyi: I'll test about it, since we don't catch exception, if the credential get token with error it will throw it and the pipeline will be terminated.

weidongxu-microsoft: The point here is that I do not want to write to the cache unless it is a success (+ token not null).

ChenTanyi: https://stackoverflow.com/questions/55976963/is-there-a-difference-between-doonsuccess-vs-doonnext-for-a-mono As SO said, it seems doOnNext is more likely to be used which would skip when data is null.

yaohaizh:
@ChenTanyi if this is the case for bear token, could we use inheritance instead? Two same names is confusing unless it is must have to use the same name?

ChenTanyi:
@yaohaizh Since the functions in base bear token policy cannot be reused, so I think the inheritance is useless. The name is used like UserAgentPolicy for using same name. Do we need to change it into some name like `DefaultScopeAuthenticationPolicy`?

PR #1004, opened 2020-02-05
feat: migrate resource group tagging interceptor to V2

ChenTanyi:


azure-mgmt-resources/src/test/java/com/azure/management/resources/core/ResourceGroupTaggingPolicy.java
```diff
@@ -0,0 +1,88 @@
...
+                byteBuffer -> {
+                    byte[] body = new byte[byteBuffer.remaining()];
+                    byteBuffer.get(body);
+                    String bodyStr = new String(body, StandardCharsets.UTF_8);
+
+                    ResourceGroupInner resourceGroupInner;
+                    try {
+                        resourceGroupInner = adapter.deserialize(bodyStr, ResourceGroupInner.class, SerializerEncoding.JSON);
+                        if (resourceGroupInner == null) {
+                            throw new IOException();
```
weidongxu-microsoft: Do not use exception as flow control.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/ResourceGroupTaggingPolicy.java
```diff
@@ -0,0 +1,88 @@
+
+
```
weidongxu-microsoft: Clean up a bit.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/ResourceGroupTaggingPolicy.java
```diff
@@ -0,0 +1,88 @@
...
+                        return Mono.error(new RuntimeException("Failed to deserialize " + bodyStr));
+                    }
+
+                    Map<String, String> tags = resourceGroupInner.getTags();
+                    if (tags == null) {
+                        tags = new HashMap<>();
+                    }
+                    tags.put("product", "javasdk");
+                    tags.put("cause", "automation");
+                    tags.put("date", Instant.now().toString());
```
weidongxu-microsoft: Will it overwrite users with same name?

ChenTanyi:
@weidongxu-microsoft The V1 interceptor is just the deletion file.
Or you can see it at https://github.com/azure/azure-libraries-for-java/blob/7ef6e15387864cf3b41961583b580d8f570ce9aa/azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/core/ResourceGroupTaggingInterceptor.java

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/ResourceGroupTaggingPolicy.java
ChenTanyi: Yes, so it is only used at unit test.

ChenTanyi:
@weidongxu-microsoft It adds tags only for unit test, so may not be suitable to put in the ResourceGroupImpl.

weidongxu-microsoft:
Then you are free to go if it is only for testing.

Please still fix the exception part. One should not use exception for normal if/else handling.

ChenTanyi:
> Then you are free to go if it is only for testing.
> 
> Please still fix the exception part. One should not use exception for normal if/else handling.

Done

PR #1005, opened 2020-02-05
Migrate Auxiliary Authentication Policy to V2

ChenTanyi:
Migration from V1 interceptor.
V1: https://github.com/azure/azure-libraries-for-java/blob/bc5a8263bb43d6f03ef6cc74d99efe8f1c925b29/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/AuxiliaryCredentialsInterceptor.java#L24

Used for cross tenant authentication, no test yet. I may test it later and will not merge it until my test succeed.


azure-mgmt-resources/src/main/java/com/azure/management/AzureTokenCredential.java
```diff
@@ -43,42 +44,33 @@ public AzureTokenCredential(AzureEnvironment environment, String domain) {
...
-
-    /**
-     * Override this method to provide the mechanism to get a token.
-     *
-     * @param resource the resource the access token is for
-     * @return the token to access the resource
-     * @throws IOException exceptions from IO
-     */
-//    public abstract Mono<AccessToken> getToken(String resource);
+    public final Mono<AccessToken> getToken(HttpRequest request) {
```
weidongxu-microsoft: Could you do a proper refresh? These code seems not belong here.

ChenTanyi: I move it to here due to I need cached in every credential, in the second commit.

ChenTanyi:
Test passed.
* doc: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/authenticate-multi-tenant
* example: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/share-images-across-tenants, https://medium.com/@ArsenVlad/azure-vnet-peering-across-azure-active-directory-tenants-using-service-principal-authentication-13c52d3190ab

azure-mgmt-resources/src/main/java/com/azure/management/BearerTokenAuthenticationPolicy.java
```diff
@@ -76,21 +35,11 @@ private String getDefaultScopeFromRequest(HttpRequest request) {
...
         Mono<AccessToken> tokenResult;
-        AccessToken token = tokenCache.get(scopes[0]);
-        if (token == null || token.isExpired()) {
-            tokenResult = this.credential.getToken(new TokenRequestContext().addScopes(scopes))
-                            .doOnNext(accessToken -> this.tokenCache.put(scopes[0], accessToken));
+        if (this.scopes == null || this.scopes.length == 0) {
+            tokenResult = this.credential.getToken(context.getHttpRequest());
         } else {
-            tokenResult = Mono.just(token);
+            tokenResult = this.credential.getToken(new TokenRequestContext().addScopes(scopes));
```
weidongxu-microsoft: After multiple PR updates, It is now a bit messed up on the scope vs. request (host).

There is 3 thing happening here:
(1) If scope is not provided, deduct the scope from host+environment
(2) Get the token based on scope
(3) Cache the token

If we stick with scope, my suggestion:
(2) is always done in ApplicationTokenCredential. No problem.
(1) is moved a lot in various PRs. My suggestion is to put it to a Util class. Input host+environment, output scope. Deterministic, and easy to do unit test. No need to couple this to either TokenCredential or Policy.
After it is decoupled, Policy could then call TokenCredential.getToken with TokenRequestContext (use the Util above to get the scope if not provided).
(3) seems better to be in subclass of TokenCredential, since AuxiliaryAuthenticationPolicy would also use that cache.

@yaohaizh Let us know your opinion.

ChenTanyi: From my perspective, I first use the TokenCredential for authorization, so every function should write outside the credential. But during the auxiliary authentication, I found that all projects are based on AzureTokenCredential, I think this common function could be written inside the AzureTokenCredential. 

Do we have scenario to use the TokenCredential rather than AzureTokenCredential?

weidongxu-microsoft: My consideration is, here you call Token on 2 cases, one is scope, another is request. So Token class got depends on two things, scope object or request object, based on things outside.

If we stick to scope, then scope object is a must. However request object is not a must have, since any static function could calculate scope from host+environment (a function which not related to Token class at all, since we just moved it multiple times now), then just provide scope to Token, which in turn make Token class only depends on scope object.

Anyway, it is up to you.

ChenTanyi: Yeah, the calculation could use a Utils to calculate (I would fix it). And the cache, I think it couldn't be based on the TokenCredential, which actually makes all credential check harder. I'll see if it is better to use a cache token in the AzureTokenCredential.

weidongxu-microsoft: Cache should be fine at current state.

Do we now still need AzureTokenCredential in RestClient? Or TokenCredential is enough?
Ask because found this code in track1, the FunctionCredentials extends TokenCredenttial
```
            functionService = manager().restClient().newBuilder()
                    .withBaseUrl(defaultHostName.toString())
                    .withCredentials(new FunctionCredentials(this))
                    .withLogLevel(LogLevel.BODY_AND_HEADERS)
                    .build()
                    .retrofit().create(FunctionService.class);
```

Not a big issue, I can handle it later when doing for appservice.

ChenTanyi: I'll check it, since the tokenCredentials in track 1 is different from track 2

ChenTanyi: I have revert the AzureTokenCredential commit, change it back to TokenCredential in restClient, and split the tokencache with lock

weidongxu-microsoft:
@ChenTanyi 

Another thing occurs to me. The `synchronized` probably will not protect the concurrency, since it is not async.

So at least you need to check if you should use ConcurrentMap on cache.

ChenTanyi:
@weidongxu-microsoft I have change token cache to ConcurrentMap already.

weidongxu-microsoft:
@ChenTanyi 
Good. Though the Lock might not work much, since it is on method, but request is done async. If 2 request come in same time, one will lock, start async, unlock, before async done the cache, another will lock, find token=null and start async, unlock.

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/AuxiliaryAuthenticationPolicy.java
```diff
@@ -0,0 +1,90 @@
+/**
```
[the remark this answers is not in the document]
ChenTanyi: The simple token cache in azure core only support for one token and can not change token acquired function. I haven't seen token cache as a class in azure identity.

ChenTanyi:
I finally didn't use the lock, since it cannot unlock in mono due to different thread problem. And use the SimpleTokenCache in azure-core instead.

weidongxu-microsoft:
Yeah, lock is not much compatible with async since it more or less based on thread. And basically one is not supposed to *wait* in async.

Something like `AtmoicBoolean.compareAndSet` could still be used, if just need to check a statue.

ChenTanyi:
@weidongxu-microsoft Yeah, the `AtmoicBoolean` is used in the SimpleTokenCache, so I just reuse this basic class.

PR #1006, opened 2020-02-06
Generate network 2019-11 and add support for PublicIPPrefixes

yaohaizh:


azure-mgmt-network/src/main/java/com/microsoft/azure/management/network/PublicIPPrefix.java
```diff
@@ -0,0 +1,213 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ * <p>
+ * Code generated by Microsoft (R) AutoRest Code Generator.
```
weidongxu-microsoft: Better to delete this line, else next generation this file will be deleted.

azure-mgmt-network/src/main/java/com/microsoft/azure/management/network/implementation/PublicIPPrefixImpl.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Override
+    public ProvisioningState provisioningState() {
+        return this.inner().provisioningState();
+    }
+
+    @Override
+    public List<ReferencedPublicIpAddress> publicIPAddresses() {
+        return this.inner().publicIPAddresses();
```
weidongxu-microsoft: Do we need to use Collections.unmodifiableList?

PR #1012, opened 2020-02-07
fix: authorization-auxiliary use "," to separate instead of ";"

ChenTanyi:
by https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/authenticate-multi-tenant


weidongxu-microsoft:
I am not really sure about it...

https://github.com/Azure/azure-cli/blob/master/src/azure-cli-core/azure/cli/core/adal_authentication.py

https://github.com/MicrosoftDocs/azure-docs/pull/28111/files#diff-3d968b87e7c00fc267607c2d72b70698

ChenTanyi:
I've tested with failure when using ';', so I change it. I'm not sure if previously it use ';'.
And actually cross tenant authentication is hardly used, especially for 3 tenants (need to separate auxiliary). 

The last change is https://github.com/MicrosoftDocs/azure-docs/pull/36669

PR #1013, opened 2020-02-09
Migrate ProviderRegistrationPolicy and ResourceManagerThrottlingPolicy

ChenTanyi:
All had been tested.
V1: https://github.com/Azure/azure-libraries-for-java/blob/bc5a8263bb43d6f03ef6cc74d99efe8f1c925b29/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ProviderRegistrationInterceptor.java
https://github.com/Azure/azure-libraries-for-java/blob/bc5a8263bb43d6f03ef6cc74d99efe8f1c925b29/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ResourceManagerThrottlingInterceptor.java

Issue: The ResourceManagerThrottling makes logging more and more, will test why

### DIFF
--- a/azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ProviderRegistrationPolicy.java
+++ b/azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ProviderRegistrationPolicy.java
@@ -1,0 +1,103 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.azure.management.resources.fluentcore.utils;
+
+import com.azure.core.http.HttpPipelineCallContext;
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpRequest;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
+import com.azure.core.management.CloudError;
+import com.azure.core.management.serializer.AzureJacksonAdapter;
+import com.azure.core.util.FluxUtil;
+import com.azure.core.util.serializer.SerializerEncoding;
+import com.azure.management.AzureTokenCredential;
+import com.azure.management.RestClient;
+import com.azure.management.RestClientBuilder;
+import com.azure.management.resources.Provider;
+import com.azure.management.resources.fluentcore.arm.ResourceUtils;
+import com.azure.management.resources.implementation.ResourceManager;
+import reactor.core.publisher.Mono;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A Http Pipeline Policy for automatic provider registration in Azure.
+ */
+public class ProviderRegistrationPolicy implements HttpPipelinePolicy {
+    private final static String MISSING_SUBSCRIPTION_REGISTRATION = "MissingSubscriptionRegistration";
+    private final AzureTokenCredential credential;
+
+    /**
+     * Initialize a provider registration policy with a credential that's authorized
+     * to register the provider.
+     * @param credential the credential for provider registration
+     */
+    public ProviderRegistrationPolicy(AzureTokenCredential credential) {
+        this.credential = credential;
+    }
+
+    private boolean responseSuccessful(HttpResponse response) {
+        return response.getStatusCode() >= 200 && response.getStatusCode() < 300;
+    }
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        final HttpRequest request = context.getHttpRequest();
+        return next.process().flatMap(
+            response -> {
+                if (!responseSuccessful(response)) {
+                    HttpResponse bufferedResponse = response.buffer();
+                    return FluxUtil.collectBytesInByteBufferStream(bufferedResponse.getBody()).flatMap(
+                        body -> {
+                            String bodyStr = new String(body, StandardCharsets.UTF_8);
+
+                            AzureJacksonAdapter jacksonAdapter = new AzureJacksonAdapter();
+                            CloudError cloudError;
+                            try {
+                                cloudError = jacksonAdapter.deserialize(bodyStr, CloudError.class, SerializerEncoding.JSON);
+                            } catch (IOException e) {
+                                return Mono.just(bufferedResponse);
+                            }
+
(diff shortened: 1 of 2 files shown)


azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ProviderRegistrationPolicy.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Initialize a provider registration policy with a credential that's authorized
+     * to register the provider.
+     * @param credential the credential for provider registration
+     */
+    public ProviderRegistrationPolicy(AzureTokenCredential credential) {
+        this.credential = credential;
+    }
+
+    private boolean responseSuccessful(HttpResponse response) {
```
yaohaizh: For boolean feauntion, should naming with `is`

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ResourceManagerThrottlingPolicy.java
```diff
@@ -0,0 +1,132 @@
...
+        String extractSubscriptionId = ResourceUtils.extractFromResourceId(request.getUrl().getPath(), "subscriptions");
+        if (extractSubscriptionId == null) {
+            extractSubscriptionId = "global";
+        }
+        final String subscriptionId = extractSubscriptionId;
+        REENTRANT_LOCK_MAP.putIfAbsent(subscriptionId, new ReentrantLock());
+        try {
+            synchronized (REENTRANT_LOCK_MAP.get(subscriptionId)) {
+                if (REENTRANT_LOCK_MAP.get(subscriptionId).isLocked()) {
+                    REENTRANT_LOCK_MAP.get(subscriptionId).wait();
```
weidongxu-microsoft: Please double check the code. I think v1 code is incorrect. Commonly no one calls `wait` on an ReentrantLock.

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ProviderRegistrationPolicy.java
```diff
@@ -0,0 +1,103 @@
...
+                            CloudError cloudError;
+                            try {
+                                cloudError = jacksonAdapter.deserialize(bodyStr, CloudError.class, SerializerEncoding.JSON);
+                            } catch (IOException e) {
+                                return Mono.just(bufferedResponse);
+                            }
+
+                            if (cloudError != null && MISSING_SUBSCRIPTION_REGISTRATION.equals(cloudError.getCode())) {
+                                String subscriptionId = ResourceUtils.extractFromResourceId(request.getUrl().getPath(), "subscriptions");
+                                RestClient restClient = new RestClientBuilder()
```
weidongxu-microsoft: Not sure about proxy, or any other configure with origin restClient.

If you are also not sure, add a TODO?

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ProviderRegistrationPolicy.java
```diff
@@ -0,0 +1,103 @@
...
+                        }
+                    );
+                }
+                return Mono.just(response);
+            }
+        );
+    }
+
+    private Provider registerProvider(String namespace, ResourceManager resourceManager) {
+        return resourceManager.providers().register(namespace);
```
weidongxu-microsoft: You seems calling a sync version? Should use the async?

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ResourceManagerThrottlingPolicy.java
ChenTanyi: `wait` makes thread waiting until `notifyAll`, it is not the method for ReentrantLock. Maybe change  it into `Object.wait()`?

weidongxu-microsoft: If one is using high level sync like ReentrantLock, one should never call low level sync method in Object (as well as synchronized block).

Please review what this class actually want to do. Normally one of lock or tryLock and then unlock should be enough to do the job.

ChenTanyi: Since it is just wait the lock to be released (after sleep for retry), but don't need to lock again, do you mean it should be better to use `lock` then `unlock` immediately?

weidongxu-microsoft: If there is no better way you can do the `wait` on the object.

What I mean is that you need to think what is the proper way to implement the feature. And what it the proper way to test to make sure it is done correctly.

Having lock inside async method is already weird enough (and the unlock is in finally block of the method, does it happen before or after the request is complete?).
1st, is there another way to throttle it in async?
2nd, if you had to do the lock, is there a better way?

@yaohaizh

ChenTanyi: I'd removed the lock, and deal with every 429 separately, since every subscription has several limitation (read, write, delete and so on)

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ProviderRegistrationPolicy.java
ChenTanyi: Done

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/ProviderRegistrationPolicy.java
ChenTanyi: Done

weidongxu-microsoft:
Nice work. Now much better.

I guess part of the intention of the original Lock is to force all requests from one subscription into a single pipeline, if there is 429 in retry (meaning if user get one 429 and that request is in retry, then even user concurrently make 10 calls, they are processed 1 by 1, no async at all). This seems not here. But actually I am not really sure whether this is the intention or not. And even it is a yes, how we can do that in async.

ChenTanyi:
I'm not sure why they force to all sync, since different operation has different limitation in one subscription. If it is a "yes", my first choice may be still the lock.

PR #1014, opened 2020-02-10
Fix compilation for tests.

yaohaizh:


azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestUtilities.java
```diff
@@ -35,4 +37,8 @@ public static void sleep(int milliseconds, boolean isRecordMode) {
         }
         return res;
     }
+
+    public static <T> boolean isEmpty(PagedIterable<T> iterable) {
```
weidongxu-microsoft: Would better to use just `Iterable`? More general type.

yaohaizh: Well, I thinking of provide a PagedList for this in future to provide back compatibility.

PR #1020, opened 2020-02-12
vnext keyvault tests

weidongxu-microsoft:
Still have 2 failure.

1. vault.update()...apply() fails due to core-mgmt polling issue (happen when get 200 directly without polling)
2. keys.list() fails, need update to 1.3 core

@ChenTanyi 
I need to remove trailing "/" in resource before add the scope for vault. Please see if it affect other types (currently I add it directly to all types, assuming they are all URLs).


ChenTanyi:
@weidongxu-microsoft I'd tested with management and graph, it's OK to remove the duplicate slash.

PR #1023, opened 2020-02-13
vnext re-generate services

weidongxu-microsoft:


weidongxu-microsoft:
@yaohaizh 

Only update on gulpfile.js is needed to enable generate client method without non-required parameters.

I regenerate network as verification. Build pass (now only need minor change on NetworkManagementClientImpl). Let me know whether you need it or not.

weidongxu-microsoft:
Re-generate multiple services, anyway I am testing latest modelerfour.

Storage tests recording included. All pass. (However I should use 2019-06

api-specs.json
```diff
@@ -21,7 +21,7 @@
     "dir": "azure-mgmt-keyvault",
     "source": "specification/keyvault/resource-manager/readme.md",
     "package": "com.azure.management.keyvault",
-    "args": "--payload-flattening-threshold=1 --tag=package-2018-02"
+    "args": "--payload-flattening-threshold=1 --tag=package-2018-02 --track2-naming=true"
```
yaohaizh: For track 2(vnext) branch, this should be the default option.

weidongxu-microsoft: OK. I will change the option to track1-naming.

PR #1029, opened 2020-02-17
refactor: rename the authentication policy and move policies into policy folder

ChenTanyi:


azure-mgmt-resources/src/test/java/com/azure/management/resources/core/AzureTestCredentials.java
```diff
@@ -20,7 +20,7 @@
     boolean isPlaybackMode;
 
     public AzureTestCredentials(final String mockUrl, String mockTenant, boolean isPlaybackMode) {
-        super("", mockTenant, "", new AzureEnvironment(new HashMap<String, String>() {{
+        super("test", mockTenant, "test", new AzureEnvironment(new HashMap<String, String>() {{
```
yaohaizh: Will this fix the PLAYBACK mode issue?

yaohaizh: Also, rename this file to AzureTestCredential. There should be no more `s` in the credential.

ChenTanyi: Yes, and actually for playback mode, it will no need for AzureTestCredentials.

ChenTanyi: Done

PR #1031, opened 2020-02-18
Upgrade junit4 to junit5

ChenTanyi:
Api changed: https://junit.org/junit5/docs/5.4.0-RC2/user-guide/index.html#migrating-from-junit4-tips
Concurrency: 32 threads per module, could be changed in properties

After I run record, it seems the providers.getByName will return null


ChenTanyi:
It seems there is issue with playback for multi-thread.

PR #1033, opened 2020-02-19
vnext paged response

weidongxu-microsoft:


azure-mgmt-appservice/src/main/java/com/azure/management/appservice/implementation/WebAppsImpl.java
```diff
@@ -82,10 +65,23 @@ protected WebAppImpl wrapModel(SiteInner inner) {
...
 
-    protected PagedIterable<WebApp> wrapList(PagedIterable<SiteInner> pagedList) {
-        return pagedList.mapPage(this::wrapModel);
-    }
+//    protected PagedIterable<WebApp> wrapList(PagedIterable<SiteInner> pagedList) {
+//        return pagedList.mapPage(this::wrapModel);
+//    }
 
+    @Override
+    protected PagedFlux<WebApp> wrapPageAsync(PagedFlux<SiteInner> innerPage) {
```
yaohaizh: Why not use PagedFlux.mapPage()

azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/utils/PagedConverter.java
```diff
@@ -0,0 +1,64 @@
...
+
+    }
+
+    /**
+     * Applies flatMap transform to elements of PagedFlux.
+     *
+     * @param mapper the flatMap transform of element T to Publisher of S.
+     * @return the PagedFlux.
+     */
+    public static <T, S> PagedFlux<S> flatMapPage(PagedFlux<T> pagedFlux, Function<? super T, ? extends Publisher<? extends S>> mapper) {
```
yaohaizh: PagedFlux.mapPage

azure-mgmt-appservice/src/main/java/com/azure/management/appservice/implementation/WebAppsImpl.java
weidongxu-microsoft: I need flatMap, not map.

PR #1035, opened 2020-02-19
Fix resources unit test with parallel

ChenTanyi:
Please: remove static variable in other package, and move all naming to the front of the test.


azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestResourceNamer.java
```diff
@@ -28,7 +29,11 @@ public TestResourceNamer(String name, InterceptorManager interceptorManager) {
     @Override
     public String randomName(String prefix, int maxLen) {
         if (interceptorManager.isPlaybackMode()) {
-            return interceptorManager.popVariable();
+            try {
```
weidongxu-microsoft: Probably better to just let it fail? Seems should not happen in normal case.

ChenTanyi: It happened usually, since we only record the variable in main thread. Multi-thread behaves strange.

ChenTanyi: If the name doesn't match, it would cause other exception in playback.

weidongxu-microsoft: I mean Exception is expected if something went wrong with recording. Better to just fail fast.

ChenTanyi: Some of the reason about it:
1. The random uuid in the async function does not affect the playback when it differ.
2. The async function would run into random thread, which could not make every record success. This makes it hard to distinguish record successful or not. 

So I just skip this error in list, if the variable affect the playback process, it would throw not match exception during request.

PR #1041, opened 2020-02-21
vnext appservice fix2

weidongxu-microsoft:
for now use BASIC AUTH for Kudu.

remove okhttp from test dependency.


weidongxu-microsoft:
Ready to be merged.

azure-mgmt-appservice/src/main/java/com/azure/management/appservice/ManagedServiceIdentity.java
```diff
@@ -14,7 +14,7 @@
  * The ManagedServiceIdentity model.
  */
 @Fluent
-public final class ManagedServiceIdentity {
+public class ManagedServiceIdentity {
```
xccc-msft: Do the tests pass when JsonInclude is not added for the map? At Compute side, I have to add that.

weidongxu-microsoft: Yes it passes.

I think for JsonInclude it more likely for a bug on service. Service must have this field even when it is nothing there. So JsonInclude there to ensure that. Unless the field is marked as `required`, I see no reason for service to do this.

PR #1044, opened 2020-02-24
Fix security issue.

yaohaizh:


yaohaizh:
Guava 21 and after are not compatible with Java 7. Close this as is. Speed up V2 development.

PR #1045, opened 2020-02-24
TestFramework: verify hostUrl in PLAYBACK mode

xccc-msft:


azure-mgmt-resources/src/test/java/com/azure/management/resources/core/InterceptorManager.java
```diff
@@ -343,4 +345,9 @@ public String popVariable() {
             return recordedData.getVariables().remove();
         }
     }
+
+    private boolean isExpectedUrl(String hostUrl, String incomingUrl) {
+        Pattern pattern = Pattern.compile("roleAssignments\\/\\w+-\\w+-\\w+-\\w+-\\w+\\?");
```
ChenTanyi: Is it due to the resource namer hasn't catched it?

xccc-msft: It's not created by resource namer. This name was generated from UUID inside the impls.

ChenTanyi: Can you link the impl here?

xccc-msft: Please refer [this line](https://github.com/Azure/azure-libraries-for-java/blob/vnext/azure-mgmt-graph-rbac/src/main/java/com/azure/management/graphrbac/implementation/RoleAssignmentHelper.java#L92)

azure-mgmt-graph-rbac/src/test/java/com/azure/management/graphrbac/GraphRbacManagementTest.java
```diff
@@ -24,7 +26,14 @@
 
     @Override
     protected void initializeClients(RestClient restClient, String defaultSubscription, String domain) {
-        graphRbacManager = GraphRbacManager.authenticate(restClient, domain);
+        RestClient graphRestClient;
+        if (isPlaybackMode()) {
+            ApplicationTokenCredential credentials = new AzureTestCredential(playbackUri.replace("http", "https"), ZERO_TENANT, true);
```
weidongxu-microsoft: `http://` to `https://` might be better against accidental match?

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/InterceptorManager.java
weidongxu-microsoft: This more like a hack to me.

If this is caused by parallel, personally I would prefer just do this case non-parallel (does JUnit has flag about this), instead of pattern matching here.

If it is not about parallel, ignore me.

azure-mgmt-graph-rbac/src/test/java/com/azure/management/graphrbac/GraphRbacManagementTest.java
xccc-msft: I'm planning to refactor the constructor of GraphRbacManager/MsiManager. If we apply same constructor like that in KeyVaultManager to pass parameters by `RestClient`, `tenantId`, `subscriptionId` instead of only `RestClient` and `tenantId`. After that, we can remove this block

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/InterceptorManager.java
ChenTanyi: @weidongxu-microsoft I think it cannot work by just non-parallel running. If it is the problem of naming during multi-thread request, it can't be resolved with the ThreadLocal namer factory.

ChenTanyi: @xccc-msft I think if we can add a function with role assignment name to solve it first.

weidongxu-microsoft: I see. It is caused by code change for parallel, but even if in serial run it still have same problem as parallel run.

Alright, without fix on root we probably had to bear with it. It could be more.

xccc-msft: @ChenTanyi @weidongxu-microsoft The problem comes from when storing the data and writing to record json. I will revert this logic and try fixing in record side.

xccc-msft:
Close this as we don't want to hack this, will find out a way to solve the issue in playback mode.

PR #1049, opened 2020-02-25
Compute: append storage account id when create option is import

xccc-msft:


azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/implementation/SnapshotImpl.java
```diff
@@ -325,4 +331,17 @@ public SnapshotImpl withSku(SnapshotSkuType sku) {
     protected Observable<SnapshotInner> getInnerAsync() {
         return this.manager().inner().snapshots().getByResourceGroupAsync(this.resourceGroupName(), this.name());
     }
+
+    private String buildStorageAccountId(String vhdUrl) {
```
yaohaizh: Naming consistent: build vs. construct. Make thing consistent is important coding convention.

azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/implementation/SnapshotImpl.java
```diff
@@ -247,7 +252,8 @@ public SnapshotImpl withDataFromVhd(String vhdUrl) {
                 .withCreationData(new CreationData())
                 .creationData()
                 .withCreateOption(DiskCreateOption.IMPORT)
-                .withSourceUri(vhdUrl);
+                .withSourceUri(vhdUrl)
+                .withStorageAccountId(buildStorageAccountId(vhdUrl));
```
weidongxu-microsoft: Just checking, if user call another `withStorageAccountId`, it will override this, right?

Since as javadoc, the actual storage account could in a different subscription.

xccc-msft: This is the enhancement we should plan. In that case, we might have new methods like `withDataFromVhd(String vhdUrl, string storageAccountId)`.

weidongxu-microsoft: I think you need to add this.

PR #1050, opened 2020-02-26
breaking change, remove Wildfly since webapp support will end near April

weidongxu-microsoft:


weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1036

PR #1051, opened 2020-02-26
vnext appservice remove guava

weidongxu-microsoft:


azure-mgmt-appservice/src/test/java/com/azure/management/appservice/implementation/UtilsTests.java
```diff
@@ -40,4 +43,26 @@ public void testWebAppPrivateRegistryImage() throws Exception {
         // completion not happen due to incorrect serviceUrl
         Assertions.assertEquals("az-func-java:v1", Utils.smartCompletionPrivateRegistryImage("az-func-java:v1", "weidxuregistry.azurecr.io"));
     }
+
+    @Test
+    public void testBase16() {
```
weidongxu-microsoft: copied from guava BaseEncodingTest.java to make sure modified base16 encode method works the same.

PR #1052, opened 2020-02-26
GraphRbac: update for playback mode

xccc-msft:


azure-mgmt-graph-rbac/src/main/java/com/azure/management/graphrbac/implementation/GraphRbacManager.java
```diff
@@ -124,14 +140,24 @@ private GraphRbacManager(RestClient restClient, String tenantId) {
...
+                .subscriptionId(subscriptionId)
                 .build();
         this.tenantId = tenantId;
     }
 
+    private String subscriptionIdFromRestClient(RestClient restClient) {
+        if (restClient.getCredential() != null && restClient.getCredential() instanceof AzureTokenCredential) {
+            return ((AzureTokenCredential) restClient.getCredential()).getDefaultSubscriptionId();
+        }
+        throw new IllegalArgumentException("The credential in RestClient must be an instance of AzureTokenCredential.");
```
ChenTanyi: The AzureTokenCredential's default subs is null, I think return null in subs may be better than throw exception.

xccc-msft: The question is whether we should inform user about this cast failure. If not mandatory, we can just return null instead of throwing exception.

ChenTanyi: If a user uses restclient directly rather than the credential in authentication, I think they already know what they want.

PR #1055, opened 2020-02-27
Add Sql in V2, remaining some test error

ChenTanyi:
@yaohaizh The canCRUDSqlSyncGroup and canCRUDSqlSyncMember have error with LRO


azure-mgmt-sql/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+      <artifactId>azure-mgmt-resources</artifactId>
+      <version>2.0.0-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-mgmt-storage</artifactId>
+      <version>2.0.0-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>com.microsoft.azure</groupId>
```
yaohaizh: Put his on the parent pom for version management. Normally, we should not put version on the child module. All version should be in the parent's dependencyManagement session. cc @xccc-msft , @weidongxu-microsoft

azure-mgmt-sql/src/main/java/com/azure/management/sql/implementation/SqlDatabaseExportRequestImpl.java
```diff
@@ -0,0 +1,170 @@
...
+            this.inner = new ExportRequest();
+        }
+        this.inner.withStorageUri(storageUri);
+        return this;
+    }
+
+    private Mono<Indexable> getOrCreateStorageAccountContainer(final StorageAccount storageAccount, final String containerName, final String fileName, final FunctionalTaskItem.Context context) {
+        final SqlDatabaseExportRequestImpl self = this;
+        return storageAccount.getKeysAsync()
+            .flatMap(storageAccountKeys -> Mono.justOrEmpty(storageAccountKeys.stream().findFirst()))
```
xccc-msft: Might use Flux.fromIterable(storageAccountKeys).singleOrEmpty() better.

azure-mgmt-sql/pom.xml
ChenTanyi: I'll change this v1 dependency to v2 dependency after the test case passed. And then add the version to parent pom.

azure-mgmt-sql/src/main/java/com/azure/management/sql/implementation/SqlDatabaseImpl.java
```diff
@@ -0,0 +1,932 @@
...
+            } else {
+                return new SqlWarehouseImpl(this.resourceGroupName, this.sqlServerName, this.sqlServerLocation, this.name(), this.inner(), this.sqlServerManager);
+            }
+        }
+
+        return null;
+    }
+
+    @Override
+    public List<RestorePoint> listRestorePoints() {
```
xccc-msft: Not sure if we plan to keep consistency of the return types with PagedIterable/PagedFlux for all modules.

azure-mgmt-sql/src/main/java/com/azure/management/sql/implementation/SqlDatabaseImpl.java
```diff
@@ -0,0 +1,932 @@
...
+    @Override
+    public Mono<TransparentDataEncryption> getTransparentDataEncryptionAsync() {
+        final SqlDatabaseImpl self = this;
+        return this.sqlServerManager.inner()
+            .transparentDataEncryptions().getAsync(this.resourceGroupName, this.sqlServerName, this.name())
+            .map(transparentDataEncryptionInner -> new TransparentDataEncryptionImpl(self.resourceGroupName, self.sqlServerName, transparentDataEncryptionInner, self.sqlServerManager));
+    }
+
+    @Override
+    public Map<String, ServiceTierAdvisor> listServiceTierAdvisors() {
```
xccc-msft: Just share sample to [convert PagedFlux to map](https://github.com/Azure/azure-libraries-for-java/blob/vnext/azure-mgmt-compute/src/main/java/com/azure/management/compute/implementation/VirtualMachineExtensionsImpl.java#L55-L60). This is optional.

PR #1057, opened 2020-02-27
feat: add test case use netty with system proxy

ChenTanyi:


azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestBase.java
```diff
@@ -224,6 +234,44 @@ public void afterTest() throws IOException {
...
+                        }
+                    }
+                }
+                String host = null;
+                int port = 0;
+                if (System.getProperty("https.proxyHost") != null && System.getProperty("https.proxyPort") != null) {
+                    host = System.getProperty("https.proxyHost");
+                    port = Integer.parseInt(System.getProperty("https.proxyPort"));
+                } else if (System.getProperty("http.proxyHost") != null && System.getProperty("http.proxyPort") != null) {
+                    host = System.getProperty("https.proxyHost");
```
weidongxu-microsoft: http.proxyHost?

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestBase.java
```diff
@@ -224,6 +234,44 @@ public void afterTest() throws IOException {
...
+                    port = Integer.parseInt(System.getProperty("http.proxyPort"));
+                }
+                if (host != null) {
+                    clientBuilder.proxy(new ProxyOptions(ProxyOptions.Type.HTTP, new InetSocketAddress(host, port)));
```
weidongxu-microsoft: Why http/https both map to HTTP, not HTTPS?

ChenTanyi: The proxy type has http only.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestBase.java
ChenTanyi: done

ChenTanyi:
It did work for no system proxy.

PR #1059, opened 2020-02-28
Fix test error in sql

ChenTanyi:


azure-mgmt-sql/src/main/java/com/azure/management/sql/implementation/SqlEncryptionProtectorImpl.java
```diff
@@ -13,8 +13,8 @@
 import com.azure.management.sql.SqlEncryptionProtector;
 import com.azure.management.sql.SqlServer;
 import com.azure.management.sql.models.EncryptionProtectorInner;
-import org.apache.commons.lang3.NotImplementedException;
 import reactor.core.publisher.Mono;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
weidongxu-microsoft: UnsupportedOperationException?

ChenTanyi: Done

PR #1066, opened 2020-03-02
vnext update record

weidongxu-microsoft:
1. use AzureJacksonAdapter with a customized DateTimeDeserializer to handle ISO8601 without timezone


weidongxu-microsoft:
@xccc-msft This will get faster playback https://github.com/Azure/azure-libraries-for-java/pull/1066/commits/e30e8f729493accd8db1b9845c892dbd4948c5c3

weidongxu-microsoft:
playback pass for appservice and keyvault

PR #1067, opened 2020-03-03
vnext, add context to client

weidongxu-microsoft:
Proposed implementation to add context for UserAgentPolicy

Common AzureServiceClient fill common part, OS, Java version.
Client fill mgmt client name.
Method fill api version.


weidongxu-microsoft:
Let me know your opinion before I start implement the generator.

azure-mgmt-resources/src/main/java/com/azure/management/resources/models/ResourceGroupsInner.java
```diff
@@ -268,7 +270,9 @@ public void delete(String resourceGroupName) {
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Mono<SimpleResponse<ResourceGroupInner>> getWithResponseAsync(String resourceGroupName) {
-        return service.get(this.client.getHost(), resourceGroupName, this.client.getSubscriptionId(), this.client.getApiVersion());
+        return FluxUtil
+                .withContext(context -> service.get(this.client.getHost(), resourceGroupName, this.client.getSubscriptionId(), this.client.getApiVersion(), context))
+                .subscriberContext(context -> context.putAll(FluxUtil.toReactorContext(this.client.getContext().addData("Sdk-Version", this.client.getApiVersion()))));
```
ChenTanyi: In https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy, the Sdk-Version is the version of sdk (1.31.0 etc.) instead of the api -version. For Sdk-Name, I think it should use the whole package name (azure.management.resources).

weidongxu-microsoft: Got it. Sdk-Name seems not the whole package name (storage.blobs is apparently not a package name).

Then I will leave both to be set from the builder (similar to host/subscriptionId, manager could set this part of the Context).

ChenTanyi: I've tested for azure-storage-blob, 12.4.0, it use this whole artifactId in user-agent. I'm not sure if all the data-plane teams use it in user-agent.
image

weidongxu-microsoft:
Intended for https://github.com/Azure/azure-libraries-for-java/issues/1090

Context in LRO seems not working correctly.

azure-mgmt-resources/src/main/java/com/azure/management/resources/models/ResourceGroupsInner.java
yungezz: is it necessary or ok to track machine id?

weidongxu-microsoft: I think there is no machine id? Just OS name/version, package name/version, java version.

azure-mgmt-resources/src/main/java/com/azure/management/AzureServiceClient.java
```diff
@@ -85,10 +92,31 @@ public String userAgent() {
...
+        if (!CoreUtils.isNullOrEmpty(OS_NAME)) {
+            context = context.addData("os.name", OS_NAME);
+        }
+        if (!CoreUtils.isNullOrEmpty(OS_VERSION)) {
+            context = context.addData("os.version", OS_VERSION);
+        }
+        if (sdkName == null) {
+            String packageName = this.getClass().getPackage().getName();
+            if (packageName.endsWith(".models")) {
+                sdkName = packageName.substring(0, packageName.length() - ".models".length());
```
weidongxu-microsoft: I am doing a bit of trick here.

`ComputeManagementClientImpl` from `com.azure.management.compute.models` would have Sdk-Name `com.azure.management.compute`

weidongxu-microsoft:
@ChenTanyi @xccc-msft @yungezz 

Let me know if there is anything need discussion.

weidongxu-microsoft:
As mentioned, the context for LRO is not solved yet.

yungezz:
@weidongxu-microsoft what's v1 useragent looks like? how to distinguish them in future? by version or by namespace?

azure-mgmt-resources/src/main/java/com/azure/management/AzureServiceClient.java
```diff
@@ -85,10 +92,52 @@ public String userAgent() {
...
+            }
+        }
+        context = context.addData("Sdk-Name", sdkName);
+        context = context.addData("Sdk-Version", SDK_VERSION);
```
yungezz: I remember there's limitation on lenght of userAgent at different service, such as cosmos. how long is length of our fixed useragent string?

weidongxu-microsoft: A typical one is such `User-Agent:azsdk-java-com.azure.management.resources/2.0.0 (13.0.1; Windows 10 10.0)`

Since user can set ApplicatinId, there is no limit on the length. Though we can truncate that ApplicationId is needed. @ChenTanyi 
https://github.com/Azure/azure-libraries-for-java/blob/vnext/azure-mgmt-resources/src/main/java/com/azure/management/UserAgentPolicy.java#L112-L117

track1 is not much to compare since Azure SDK gives new guideline for it.
https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy

For reference track1 `Azure-SDK-For-Java/null OS:Windows 10/10.0 MacAddressHash:9e21c3cc3783e949641d36ffd6a2e0c027267de76b62dcc951a57438fbc15f36 Java:13.0.1 (ResourceManagementClient, 2019-08-01)`

weidongxu-microsoft:
@yungezz 

In my understanding, Context need to be set to *each* REST API call. This provides finest grain for user. The challenge is to provide user other levels of control (e.g. global, per service, per cluster of requests e.g. those for doing one task).

Context is used for UserAgent, and Tracing, but it is more than that.

Here the code is in parent class of ManagementClient, so this is one shared piece for all REST API.

yungezz:
> @yungezz
> 
> In my understanding, Context need to be set to _each_ REST API call. This provides finest grain for user. The challenge is to provide user other levels of control consistently (e.g. global, per service, per cluster of requests e.g. those for doing one task requires a group of resources).
> 
> Context is used for UserAgent, and Tracing, but it is more than that.
> 
> Here the code is in parent class of ManagementClient, so this is one shared piece for all REST API.

Normally SDK could expose a setUserAgent() interface which depends on service(some service exposing setUserAgent(), some not) to append user customized userAgent to system default ones, which includes context from different layers: service, SDK, user customized etc.

The PR itself looks good, since it's our general context.

azure-mgmt-resources/src/main/java/com/azure/management/AzureServiceClient.java
yungezz: thanks Weidong for detail info. According to azure core tememetry, shouldn't sdk package name and version  set by azure.core?

yungezz: @Tanyi "DO enforce that the application ID is no more than 24 characters in length. " from azure core guideline, again, shouldn't this be done in azure.core/azure.core.mgmt?

ChenTanyi: @yungezz The azure core has a User-Agent policy, but it could only send constant UA, we have raised an issue and pr about it. But the azure-core team advise us to implement out own. Meanwhile, I think currently all code in `com.azure.management` could moved into azure-core-mgmt.

weidongxu-microsoft: Personally I do do not see how azure.core could get sdk package name and sdk version without we setting it to azure.core in the first place. Maybe I misunderstood the question.

weidongxu-microsoft:
For context and tracing, I think I am referring to these possible use cases. https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/core/azure-core-tracing-opentelemetry#examples, which is user's own tracing.

azure-mgmt-resources/src/main/java/com/azure/management/AzureServiceClient.java
yungezz: so `package-name`, `pacakge-version` isn't in the constant UA? then that makes a lot of sense. Thanks both to help clarify!

ChenTanyi: These are in the UA, but it cannot be changed between request. That 
 is not we need. We use a restclient for every package. So we need to change UA during request. https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java#L84

PR #1068, opened 2020-03-03
Compute: fix playback tests

xccc-msft:


azure-mgmt-compute/src/main/java/com/azure/management/compute/VirtualMachineIdentity.java
```diff
@@ -15,7 +15,7 @@
  * The VirtualMachineIdentity model.
  */
 @Fluent
-public final class VirtualMachineIdentity {
+public class VirtualMachineIdentity {
```
yaohaizh: Why change the final statements?

azure-mgmt-compute/src/test/java/com/azure/management/compute/VirtualMachineAvailabilityZoneOperationsTests.java
```diff
@@ -48,16 +50,41 @@ protected void cleanUpResources() {
...
+        ResourceGroup resourceGroup = resourceManager.resourceGroups().define(RG_NAME)
+                .withRegion(REGION)
+                .create();
+        //Create network security group
+        NetworkSecurityGroup nsg = networkManager.networkSecurityGroups().define(nsgName)
+                .withRegion(REGION)
+                .withExistingResourceGroup(resourceGroup)
+                .create();
+        //Create public IP address
+        PublicIPAddress pip = networkManager.publicIPAddresses().define(pipDnsLabel)
```
yaohaizh: Please keep the test case aka. example back compatible. Not sure why so much changes for this example?

azure-mgmt-compute/src/main/java/com/azure/management/compute/VirtualMachineIdentity.java
weidongxu-microsoft: Azure-core bug. This is the workaround until they fix.

azure-mgmt-compute/src/test/java/com/azure/management/compute/VirtualMachineAvailabilityZoneOperationsTests.java
weidongxu-microsoft: @xccc-msft You might want to double check with Tanyi. After SdkContext move to manager, and code use this SdkContext for random naming for inline resource creation, at least my test runs good on appservice (which previously got 2 kind of inline, StorageAccount, and AppServicePlan).

xccc-msft:
Close this as NSG change is not mandatory for tests here.

PR #1075, opened 2020-03-06
Compute: fix playback tests

xccc-msft:
all playback tests of compute/graph/msi pass.


azure-mgmt-compute/src/main/java/com/azure/management/compute/implementation/VirtualMachineScaleSetImpl.java
```diff
@@ -334,7 +334,7 @@ public Network getPrimaryNetwork() throws IOException {
     @Override
     public LoadBalancer getPrimaryInternetFacingLoadBalancer() throws IOException {
         if (this.primaryInternetFacingLoadBalancer == null) {
-            loadCurrentPrimaryLoadBalancersIfAvailable();
+            loadCurrentPrimaryLoadBalancersIfAvailable().block();
```
weidongxu-microsoft: Maybe you can name it as loadCurrentPrimaryLoadBalancersIfAvailableAsync to make it clearer that it is async method.

azure-mgmt-compute/src/main/java/com/azure/management/compute/implementation/VirtualMachineScaleSetImpl.java
```diff
@@ -1902,17 +1906,18 @@ private void loadCurrentPrimaryLoadBalancersIfAvailable() throws IOException {
...
 
-        LoadBalancer loadBalancer1 = this.networkManager
-                .loadBalancers()
-                .getById(firstLoadBalancerId);
-        if (loadBalancer1.publicIPAddressIds() != null && loadBalancer1.publicIPAddressIds().size() > 0) {
-            this.primaryInternetFacingLoadBalancer = loadBalancer1;
-        } else {
-            this.primaryInternalLoadBalancer = loadBalancer1;
-        }
+        self = self.concatWith(Mono.just(firstLoadBalancerId).flatMap(id -> this.networkManager.loadBalancers().getByIdAsync(id)
```
weidongxu-microsoft: `self.map(ignored -> firstLoadBalancerId).flatMap(id -> ...` might be better, since it keeps `Mono` (if I did not misunderstand, self is `Mono.just(this)`)

PR #1079, opened 2020-03-09
Add CI pipeline and Fix azure, azure-sample test

ChenTanyi:


ChenTanyi:
The CI has passed https://dev.azure.com/tanchentest/azure-libraries-for-java/_build/results?buildId=67&view=logs&j=572c5e49-83d5-5271-390a-e6dc77f89c6b

Have a lot of commits. Before `merge d8d2f4a` is dealing with ci yml and some simple test. After `merge` is dealing with `azure` and `azure-sample` test.

ChenTanyi:
@weidongxu-microsoft you will need to setup a pipeline after it merged.

weidongxu-microsoft:
@ChenTanyi

You can share your CI pipeline here.

New one probably not going to work, since the yaml is not on master.

azure-mgmt-network/src/main/java/com/azure/management/network/NetworkSecurityGroup.java
```diff
@@ -16,6 +16,7 @@
 import com.azure.management.resources.fluentcore.arm.models.GroupableResource;
 import com.azure.management.resources.fluentcore.arm.models.Resource;
 import com.azure.management.resources.fluentcore.model.Appliable;
+import com.azure.management.resources.fluentcore.model.Attachable;
```
weidongxu-microsoft: Is these `import` necessary, given it is already compile pass?

.azure-pipelines/fluentJavaMgmt.yml
```diff
@@ -40,41 +49,19 @@ jobs:
...
-          options: '$(ProfileFlag)'
-          mavenOptions: '$(LoggingOptions)'
-          javaHomeOption: 'JDKVersion'
-          jdkVersionOption: $(JavaVersion)
-          jdkArchitectureOption: 'x64'
-          publishJUnitResults: false
-          goals: 'jetty:run-forked'
-          
+      # - task: Maven@3
+      #   displayName: 'Run StyleCheck'
```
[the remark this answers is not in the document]
ChenTanyi: The stylecheck has many error with auto-gen code, I firstly fix the unit test, so comment it first.

ChenTanyi:
@weidongxu-microsoft My pipeline is using for test, in my link above. If there is problem for setting up pipeline in internal devops, I'll set up a new pipeline in my project first for public.

azure-mgmt-network/src/main/java/com/azure/management/network/NetworkSecurityGroup.java
ChenTanyi: It fixes a warning for https://github.com/azure/azure-libraries-for-java/blob/a3595380aabbcc6e2fa97e6f52480a6d5d0d6393/azure-mgmt-network/src/main/java/com/azure/management/network/NetworkSecurityGroup.java#L85

ChenTanyi:
@weidongxu-microsoft It seems I cannot create hook at this project.
image

azure-mgmt-compute/src/main/java/com/azure/management/compute/implementation/VirtualMachineScaleSetImpl.java
```diff
@@ -1468,14 +1468,15 @@ public VirtualMachineScaleSetImpl withoutUserAssignedManagedServiceIdentity(Stri
     // Create Update specific methods
     //
     @Override
-    protected void beforeCreating() {
+    protected Mono<Void> beforeCreating() {
```
xccc-msft: Why we are having such kind of change from void to Mono<Void>?

ChenTanyi: Because this function could be implemented with blocking, which is not allowed during creation, so I change to Mono which can be chained.

xccc-msft: But in that case, you could wrap in createResourceAsync when it's required. I doubt if it is good style to do like this. After your changes, a lot of code snippet will work as it is while it just returns Mono.empty().

ChenTanyi: Yes, but I'm afraid something bad happened if I change the blocking into `createResourceAsync`. The framework is not written by me, I can't guarantee to change everywhere. Anyway, I could try to change it. And actually, for this implementation, `createResourceAsync` is not necessary, it is splited into `beforeCreating`, `createInner` and `afterCreating` three necessary function. It could be strange when implement them all.

weidongxu-microsoft: I am not sure but from v1 code in `GroupableParentResourceImpl` it seems the intention for beforeCreating and afterCreating is do things no blocking (so no request). But we have bug in v1 that do block.

So it fails in v2 after migration.

Agree it might be better for us to fix the blocking issue in before/after if possible, as done in latest commit.

ChenTanyi: Change it back, and use non-blocking chaining in another way.

ChenTanyi:
@yungezz Can you give me the permission to add hook? I need this permission to add azure-pipeline. And it could be removed after the pipeline is added.

ChenTanyi:
The pipeline is currently set at https://dev.azure.com/azure-libraries-for-java-test/Public/_build?definitionId=1&_a=summary

weidongxu-microsoft:
@ChenTanyi 

I think I can create the pipeline after this PR is merged.
The thing I am not sure is whether I can get it run on PR targeting vnext branch, or get it auto-triggered on vnext commit. The yaml on master might take precedence,
Anyway we could try it after merge.

ChenTanyi:
They are different hooks, I think they will run together, maybe the master trigger could remove. I'll create a PR about it.

azure-mgmt-network/src/main/java/com/azure/management/network/implementation/RouteTableImpl.java
```diff
@@ -132,11 +132,6 @@ protected void beforeCreating() {
         this.inner().withRoutes(innersFromWrappers(this.routes.values()));
     }
 
-    @Override
-    protected void afterCreating() {
-        initializeChildrenFromInner();
```
weidongxu-microsoft: Is it called somewhere instead?

azure-mgmt-network/src/main/java/com/azure/management/network/implementation/VirtualNetworkGatewayImpl.java
```diff
@@ -329,11 +327,6 @@ protected void beforeCreating() {
         this.inner().withIpConfigurations(innersFromWrappers(this.ipConfigs.values()));
     }
 
-    @Override
-    protected void afterCreating() {
-        initializeChildrenFromInner();
```
weidongxu-microsoft: Is it called somewhere instead?

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestBase.java
```diff
@@ -290,7 +290,7 @@ protected String baseUri() {
         }
     }
 
-    protected RestClient buildRestClient(RestClientBuilder builder, boolean isMocked) {
+    protected synchronized RestClient buildRestClient(RestClientBuilder builder, boolean isMocked) {
```
weidongxu-microsoft: Would you let us know why it need `synchronized`. This method seems not sharing anything (and almost just a static method)?

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestUtilities.java
```diff
@@ -19,7 +19,7 @@
      *
      * @param milliseconds time in milliseconds for which to sleep.
      */
-    public static void sleep(int milliseconds, boolean isRecordMode) {
+    public synchronized static void sleep(int milliseconds, boolean isRecordMode) {
```
weidongxu-microsoft: Same question. These are static method sharing nothing?

azure-samples/src/main/java/com/azure/management/samples/Utils.java
```diff
@@ -463,8 +463,8 @@ public static void print(Vault vault) {
                 .append("\n\tAccess policies: ");
         for (AccessPolicy accessPolicy : vault.accessPolicies()) {
             info.append("\n\t\tIdentity:").append(accessPolicy.objectId())
-                    .append("\n\t\tKey permissions: ").append(Joiner.on(", ").join(accessPolicy.permissions().getKeys()))
-                    .append("\n\t\tSecret permissions: ").append(Joiner.on(", ").join(accessPolicy.permissions().getSecrets()));
+                    .append("\n\t\tKey permissions: ").append(String.join(", ", accessPolicy.permissions().getKeys().stream().map(key -> key.toString()).collect(Collectors.toList())))
```
weidongxu-microsoft: There is a `Collectors.joining` as well.

azure-mgmt-network/src/main/java/com/azure/management/network/implementation/RouteTableImpl.java
ChenTanyi: The `initializeChildrenFromInner` is call before `afterCreating` in `GroupableParentResourceImpl.createResourceAsync`, https://github.com/azure/azure-libraries-for-java/blob/95bc556/azure-mgmt-resources/src/main/java/com/azure/management/resources/fluentcore/arm/models/implementation/GroupableParentResourceImpl.java#L57

azure-mgmt-network/src/main/java/com/azure/management/network/implementation/VirtualNetworkGatewayImpl.java
ChenTanyi: same above.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestBase.java
ChenTanyi: It call a static function in `RetryPolicy`, which will cause iterator error if not synchronized.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestUtilities.java
ChenTanyi: Maybe this synchronized could be removed, I just add it to most static method, since they are dangerous in parallel test.

azure-samples/src/main/java/com/azure/management/network/samples/ManageVirtualNetworkAsync.java
```diff
@@ -131,15 +143,9 @@ public static boolean runSample(final Azure azure) {
             ).map(indexable -> {
                 if (indexable instanceof NetworkSecurityGroup) {
                     NetworkSecurityGroup nsg = (NetworkSecurityGroup) indexable;
-                    System.out.println("Created network security group");
-                    // Print the network security group
-                    Utils.print(nsg);
```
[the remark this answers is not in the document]
ChenTanyi: Blocking operation is in the `Utils.print`, it call some get function.

azure-samples/src/main/java/com/azure/management/samples/Utils.java
ChenTanyi: I'll test it later.

weidongxu-microsoft: No need to change now. Just a heads-up.

azure-samples/src/main/java/com/azure/management/network/samples/ManageVirtualNetworkAsync.java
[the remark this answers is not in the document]
weidongxu-microsoft: Got it.

azure-samples/src/main/java/com/azure/management/samples/Utils.java
ChenTanyi: I've added it for test, since it could shorten the code.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestUtilities.java
ChenTanyi: Removed some unnecessary synchronized.

PR #1091, opened 2020-03-11
Container Service: migrate to V2

ChenTanyi:


api-specs.json
```diff
@@ -99,8 +99,8 @@
   "containerservice": {
     "dir": "azure-mgmt-containerservice",
     "source": "specification/containerservice/resource-manager/readme.md",
-    "package": "com.microsoft.azure.management.containerservice",
-    "args": "--payload-flattening-threshold=1 --tag=package-2019-08 --title=ContainerServiceManagementClient --description=\"Container Service Client\""
+    "package": "com.azure.management.containerservice",
+    "args": "--payload-flattening-threshold=1 --tag=package-2019-08 --title=ContainerServiceManagementClient --description=\"Container Service Client\" --track1-naming=true"
```
yungezz: what's track1-naming mean?

azure-mgmt-containerservice/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+        <groupId>com.azure</groupId>
+        <artifactId>azure-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>azure-mgmt-containerservice</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Microsoft Azure SDK for Container Service Management</name>
```
yungezz: should we use azure kubernetes service now?

azure-mgmt-containerservice/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+    <licenses>
+        <license>
+            <name>The MIT License (MIT)</name>
+            <url>http://opensource.org/licenses/MIT</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <scm>
+        <url>scm:git:https://github.com/Azure/azure-sdk-for-java</url>
```
yungezz: why use SDK repo

azure-mgmt-containerservice/src/main/java/com/azure/management/containerservice/AgentPoolAvailableVersionsPropertiesAgentPoolVersionsItem.java
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for
+// license information.
+// 
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+package com.azure.management.containerservice;
```
yungezz: same here, should we change namespace to kubernets service?

azure-mgmt-containerservice/src/main/java/com/azure/management/containerservice/CloudError.java
```diff
@@ -0,0 +1,42 @@
...
+package com.azure.management.containerservice;
+
+import com.azure.core.annotation.Fluent;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * The CloudError model.
+ */
+@Fluent
+public final class CloudError {
```
yungezz: seems one change in azure core is to generalize exception, not using CloudError, is this fluent layer?

api-specs.json
ChenTanyi: The track1-naming is used to generate code with naming as track1, "with***" instead of "set***". cc @weidongxu-microsoft

azure-mgmt-containerservice/pom.xml
ChenTanyi: The Container Service uses the name of the provider. And it include not only kubernetes, but also container service. I think if we want to name it to kubernetes, we need to separate these two service.

azure-mgmt-containerservice/pom.xml
ChenTanyi: I'm not sure about it, all the poms are used it in this project. Meanwhile, since we will migrate back to sdk, may it be ok?

azure-mgmt-containerservice/src/main/java/com/azure/management/containerservice/AgentPoolAvailableVersionsPropertiesAgentPoolVersionsItem.java
ChenTanyi: containerservice here, I think we couldn't change. https://github.com/azure/azure-rest-api-specs/blob/1c3a60403422ca8c73df7eb7c65ae2a3c90bf152/specification/containerservice/resource-manager/readme.md#L1

azure-mgmt-containerservice/src/main/java/com/azure/management/containerservice/CloudError.java
ChenTanyi: This is a generated class, which is defined by [swagger](https://github.com/azure/azure-rest-api-specs/blob/1c3a60403422ca8c73df7eb7c65ae2a3c90bf152/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2019-08-01/managedClusters.json#L2269). They use spatial cloud error rather than common one.

weidongxu-microsoft: This could be a generator bug. CloudError and CloudException is defined in Core Management. Hence generator does not need to generate CloudError class.

azure-mgmt-containerservice/src/main/java/com/azure/management/containerservice/AgentPoolAvailableVersionsPropertiesAgentPoolVersionsItem.java
yaohaizh: @yungezz this namespace is defined in readme.md by the service team when write the Swagger spec. All the SDK will have the similar namespace. 
 
FYI: 
       `ContainerService`: AKS
       `ContainerInstance`: ACI
       `ContainerRegistry`: ACR

api-specs.json
weidongxu-microsoft: I probably going to put it to gulp.js, so this is enabled globally.

yaohaizh: Should keep track-1 as default. And add 'azure-core-style' as alternative.

PR #1093, opened 2020-03-11
migrate keyvault sample

weidongxu-microsoft:


azure-samples/pom.xml
```diff
@@ -54,31 +54,6 @@
             <artifactId>azure</artifactId>
```
weidongxu-microsoft: Sample depends on azure, hence it should not need the SDKs, since azure already have them all.

gulpfile.js
```diff
@@ -133,7 +133,7 @@ function codegen(project, cb) {
                         generator +
                         ` --namespace=${mappings[project].package} ` +
                         ` --java.output-folder=${outDir} ` +
-                        ` --license-header=MICROSOFT_MIT_NO_CODEGEN ` +
+                        ` --license-header=MICROSOFT_MIT_SMALL ` +
```
weidongxu-microsoft: Use shorter copyright as yaohai requested.

PR #1095, opened 2020-03-12
feat: add sdk context to azure in test base

ChenTanyi:


weidongxu-microsoft:
Mostly, it wraps something that need to be handled differently in live and test. E.g. test need recording so random name will not work; test need to be fast, so some sleep or time duration is reduced to 0.

ChenTanyi:
The sdkcontext is used to set some different parameters for record and playback test., which couldn't work fine for parallel test as static variable. Thus we need to set it into manager.

PR #1096, opened 2020-03-12
Migration azure and samples: sql, cosmosdb, container service

ChenTanyi:
Test not passed yet, I'll fix it together with CI-pipeline


azure-samples/pom.xml
```diff
@@ -87,6 +87,16 @@
             <groupId>com.azure</groupId>
             <artifactId>azure-storage-blob</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>azure-documentdb</artifactId>
```
weidongxu-microsoft: Please take a look whether you can use latest cosmos (not sure, azure-cosmos 4.0.0?).

azure-samples/pom.xml
```diff
@@ -87,6 +87,16 @@
...
+            <version>1.9.4</version>
+        </dependency>
+        <dependency>
+            <groupId>joda-time</groupId>
+            <artifactId>joda-time</artifactId>
```
weidongxu-microsoft: I don't think we should do this explicitly in track2 (it is ok if it from another lib, but then you will not need to write it here). Check your code and make refactor.

azure-samples/pom.xml
ChenTanyi: done, version is 4.0.1-beta.1

azure-samples/pom.xml
ChenTanyi: done

PR #1097, opened 2020-03-13
fix: proxy address error in new netty and add timeout policy

ChenTanyi:


ChenTanyi:
@weidongxu-microsoft My test is working, but not so sure. (Yesterday, I found 2 test is still running, but I haven't check if it is timeout.) I don't think we could use it immediately in restclient, you all could test about it, later. Also, timeout could be separated to connect, read and write timeout, but the policy doesn't specify it.

xccc-msft:
> Does the TimeoutPolicy work correctly? Should we put it to default (something like `withTimeout`)?

Additionally, we need to check if TimeoutPolicy works with other policy like RetryPolicy collabratively so we can decide how to expose in the future.

azure-mgmt-resources/src/test/java/com/azure/management/resources/core/TestBase.java
```diff
@@ -212,6 +214,7 @@ public void write(int b) {
                     .withHttpClient(generateHttpClientWithProxy(null))
                     .withHttpLogOptions(new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS))
                     .withPolicy(new ResourceGroupTaggingPolicy())
+                    .withPolicy(new TimeoutPolicy(Duration.ofMinutes(1)))
```
yaohaizh: Need consider several cases for timeout: 
- Hot is working on the Polling case? 
- How is it working with the server side retry logic header (429) ?

ChenTanyi: Yes, so it is just added for unit test currently, since parallel test will cause timeout usually.
In my opinion,
* The polling case could remain the same, but add logic for cancel if there are lots of time out.
* The 429 handler should be added before timeout in the product situation.

PR #1098, opened 2020-03-13
Resources: fix NPE when trying to get non-existing resource

xccc-msft:
fix [issue](https://github.com/Azure/azure-sdk-for-java/issues/8935)


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/GenericResourcesImpl.java
```diff
@@ -90,7 +90,8 @@ public boolean checkExistenceById(String id) {
     public GenericResource getById(String id) {
         Provider provider = this.manager().providers().getByName(ResourceUtils.resourceProviderFromResourceId(id));
         String apiVersion = ResourceUtils.defaultApiVersion(id, provider);
-        return wrapModel(this.inner().getById(id, apiVersion)).withApiVersion(apiVersion);
+        GenericResourceImpl genericResource = wrapModel(this.inner().getById(id, apiVersion));
+        return genericResource == null ? null : genericResource.withApiVersion(apiVersion);
```
yaohaizh: In which case the result be null?

xccc-msft: When user request to get unavailable resource(like incorrect id or resource deleted), the server return CloudException. From SDK side, we catch the exception and return NULL inside the generated inners object.

Issue #1094, opened 2020-03-11
[BUG] azure-mgmt-resources test-jar not published


weidongxu-microsoft:
SDK libraries depends on it. Without that test-jar, SDK libraries can only run compile phase, it will fail on test or package phase.

```
        <dependency>
            <groupId>com.azure</groupId>
            <artifactId>azure-mgmt-resources</artifactId>
            <version>1.31.1</version>
            <type>test-jar</type>
            <scope>test</scope>
        </dependency>
```

weidongxu-microsoft:
Ann Meng from Docs team reported this. It likely need to be solved before next release.

nickzhums:
Thanks, we need this for the new Java doc generation pipeline

PR #1114, opened 2020-03-18
move test-jar to profile

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1094

profile id is `azure-mgmt-sdk-test-jar`

### DIFF
--- a/azure-mgmt-appservice/pom.xml
+++ b/azure-mgmt-appservice/pom.xml
@@ -91,13 +91,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.microsoft.azure</groupId>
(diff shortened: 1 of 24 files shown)


weidongxu-microsoft:
Tried with `release:prepare`, the dependency version in profile will be updated as before.

ChenTanyi:
@weidongxu-microsoft Is it any other command needed in release?

weidongxu-microsoft:
@ChenTanyi 

Should be no. The profile is activated by `!maven.test.skip`, which we don't do during release, or by explicitly disable via `-P`.

For safety, Thursday let us first release it to stage to see the pom.

PR #1115, opened 2020-03-18
fix security

weidongxu-microsoft:
Bump jackson-databind from 2.9.4 to 2.9.10.3  dependencies
Bump checkstyle from 8.24 to 8.29 in /azure-mgmt-cosmosdb  dependencies


weidongxu-microsoft:
@yaohaizh 

Found this branch https://github.com/Azure/azure-libraries-for-java/compare/fix_security

It seems guava upgrade caused the failure. Should I try upgrade azure-client-authentication and azure-client-runtime?

PR #1120, opened 2020-03-19
Release 1.32.0

ChenTanyi:


azure-mgmt-appservice/pom.xml
```diff
@@ -8,7 +8,7 @@
   <parent>
     <groupId>com.microsoft.azure</groupId>
     <artifactId>azure-parent</artifactId>
-    <version>1.31.2-SNAPSHOT</version>
+    <version>1.32.1-SNAPSHOT</version>
```
yungezz: can we put a default version in parent pom?

ChenTanyi: I'm not sure about what central version means? We need to update dependency sometime in azure-parent, so it need to release a version with other packages as well.

PR #1122, opened 2020-03-19
vnext sample appservice

weidongxu-microsoft:
Record tests all pass.

Might need some work on live-only tests.

A few on trafficmanager and containerregistry is commented out.


azure-samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppStorageAccountConnection.java
```diff
@@ -0,0 +1,199 @@
...
+        BlobClient blobClient = blobContainerClient.getBlobClient(fileName);
+        File file = new File(fileName);
+        try (InputStream is = new FileInputStream(file)) {
+            blobClient.upload(is, file.length());
+        } catch (IOException e) {
+            System.out.println(e.getMessage());
+            e.printStackTrace();
+        }
+    }
+}
```
weidongxu-microsoft: @xccc-msft this is migrate to latest storage blob client. might be helpful if you need migrate in compute.

weidongxu-microsoft:
This is ready. I even fixed a bug in war package...

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1086

Issue #1084, opened 2020-03-11
[FEATURE REQ] Missing apiVersion parameter to generic resource


weidongxu-microsoft:
Blocked by modelerfour https://github.com/Azure/autorest.modelerfour/issues/183

Worst case: generated code can be manually modified, or certain option could be done to generator to do special handling.

PR #1126, opened 2020-03-23
vnext regen resources, fix api-version method parameter

weidongxu-microsoft:
fix https://github.com/Azure/azure-libraries-for-java/issues/1084


azure-mgmt-resources/src/main/java/com/azure/management/resources/AliasPathType.java
```diff
@@ -1,7 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for
-// license information.
-// 
+// Licensed under the MIT License.
```
xccc-msft: Should we apply this change to all existing files?

weidongxu-microsoft: We can get some script to do this. Maybe later when someone finished more important tasks...

Currently newly generated code will use this short style.

weidongxu-microsoft:
PR on spec https://github.com/Azure/azure-rest-api-specs/pull/8803

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1084

azure-mgmt-resources/src/main/java/com/azure/management/resources/DeploymentProperties.java
```diff
@@ -24,7 +22,8 @@
     private Object template;
 
     /*
-     * Entity representing the reference to the template.
+     * The URI of the template. Use either the templateLink property or the
```
yungezz: is this change caused by generator change also?

azure-mgmt-resources/src/main/java/com/azure/management/resources/implementation/DeploymentsImpl.java
```diff
@@ -40,8 +41,7 @@
 
     @Override
     public PagedIterable<Deployment> listByResourceGroup(String groupName) {
-        // FIXME: filter & top parameter
-        return this.manager().inner().deployments().listByResourceGroup(groupName, null, null).mapPage(inner -> createFluentModel(inner));
+        return this.manager().inner().deployments().listByResourceGroup(groupName).mapPage(inner -> createFluentModel(inner));
```
yungezz: just confirm, this is change from generator

azure-mgmt-resources/src/main/java/com/azure/management/resources/DeploymentProperties.java
weidongxu-microsoft: Yes. I think it is from changes by modelerfour. Not confirmed.

azure-mgmt-resources/src/main/java/com/azure/management/resources/implementation/DeploymentsImpl.java
weidongxu-microsoft: Actually this is Impl change.

The resources RP is among the 1st to migrate. Hence there remains quite some TODO/FIXME.

The method call is actually same, but older generate does not generate the method with default parameters (hence the 2 `null`s in code before).

Issue #1088, opened 2020-03-11
[FEATURE REQ] Keyvault client library still using Azure Core 1.2.0


weidongxu-microsoft:
Which is not compatible with Core 1.3.0

yungezz:
here client library means Dataplane client, right?

weidongxu-microsoft:
Currently keyvault live test will fail.

Previous it success when keyvault client and our lib both uses azure core 1.2.0

```
java.lang.RuntimeException: Failed to deserialize 5-parameter response without decoded headers.

	at com.azure.core.http.rest.ResponseConstructorsCache.lambda$invoke$2(ResponseConstructorsCache.java:132)
	at reactor.core.publisher.MonoDefer.subscribe(MonoDefer.java:44)
	at reactor.core.publisher.Mono.subscribe(Mono.java:4087)
	at reactor.core.publisher.FluxSwitchIfEmpty$SwitchIfEmptySubscriber.onComplete(FluxSwitchIfEmpty.java:75)
	at reactor.core.publisher.FluxMap$MapSubscriber.onComplete(FluxMap.java:136)
	at reactor.core.publisher.MonoProcessor$NextInner.onComplete(MonoProcessor.java:518)
	at reactor.core.publisher.MonoProcessor.onNext(MonoProcessor.java:308)
	at reactor.core.publisher.MonoProcessor.onComplete(MonoProcessor.java:265)
	at reactor.core.publisher.Operators.complete(Operators.java:132)
	at reactor.core.publisher.MonoEmpty.subscribe(MonoEmpty.java:45)
	at reactor.core.publisher.Mono.subscribe(Mono.java:4087)
	at reactor.core.publisher.MonoProcessor.add(MonoProcessor.java:457)
	at reactor.core.publisher.MonoProcessor.subscribe(MonoProcessor.java:370)
	at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:55)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:150)
	at reactor.core.publisher.Operators$MonoSubscriber.complete(Operators.java:1592)
	at reactor.core.publisher.MonoProcessor.onNext(MonoProcessor.java:317)
	at reactor.core.publisher.Operators$MonoSubscriber.complete(Operators.java:1592)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:144)
[...]

PR #1140, opened 2020-03-24
upgrade keyvault client libs

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1088


weidongxu-microsoft:
These are based on azure core 1.3.0

weidongxu-microsoft:
After this, live tests on keyvault is all green.

PR #1141, opened 2020-03-24
[WIP] DNS: update for track 2 migration

xccc-msft:


azure-mgmt-dns/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-http-netty</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-mgmt-resources</artifactId>
+      <version>2.0.0-SNAPSHOT</version>
```
yungezz: why this one version is not managed in parent?

azure-mgmt-dns/src/main/java/com/azure/management/dns/CloudError.java
```diff
@@ -0,0 +1,40 @@
...
+package com.azure.management.dns;
+
+import com.azure.core.annotation.Fluent;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * The CloudError model.
+ */
+@Fluent
+public final class CloudError {
```
yungezz: still CloudError?

weidongxu-microsoft: @xccc-msft You can just delete it. Generator will no longer generate this class.

azure-mgmt-dns/pom.xml
xccc-msft: No we cannot manage in parent pom as this is not external dependency.

azure-mgmt-dns/src/main/java/com/azure/management/dns/CloudError.java
xccc-msft: Deleted.

PR #1145, opened 2020-03-25
vnext regen storage

weidongxu-microsoft:
Regen storage after generator updated.


azure-mgmt-appservice/src/test/resources/session-records/canCRUDFunctionApp.json
```diff
@@ -189,7 +189,7 @@
     }
   }, {
     "Method" : "POST",
```
yungezz: just curious, what's the query param `expand=kerb` used for

azure-mgmt-storage/src/main/java/com/azure/management/storage/Action.java
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+package com.azure.management.storage;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
```
yungezz: why this new file not in preview generation? swagger change?

azure-mgmt-storage/src/main/java/com/azure/management/storage/models/BlobServicesInner.java
```diff
@@ -181,23 +184,23 @@ public BlobServicePropertiesInner setServiceProperties(String resourceGroupName,
...
+     * @param accountName The name of the storage account within the specified resource group. Storage account names must be between 3 and 24 characters in length and use numbers and lower-case letters only.
      * @throws IllegalArgumentException thrown if parameters fail the validation.
      * @throws CloudException thrown if the request is rejected by server.
      * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Mono<SimpleResponse<BlobServicePropertiesInner>> getServicePropertiesWithResponseAsync(String resourceGroupName, String accountName) {
         final String blobServicesName = "default";
-        return service.getServiceProperties(this.client.getHost(), resourceGroupName, accountName, this.client.getSubscriptionId(), blobServicesName, this.client.getApiVersion());
+        return service.getServiceProperties(this.client.getHost(), resourceGroupName, accountName, this.client.getApiVersion(), this.client.getSubscriptionId(), blobServicesName);
```
yungezz: this raises a little concern, curious why order of params changed

weidongxu-microsoft: change in modelerfour. currently there is no strict order on parameters, only non-required parameter after required parameter.

azure-mgmt-storage/src/main/java/com/azure/management/storage/Action.java
weidongxu-microsoft: This is the fix on non-required constant. This enum got only one value. But since it is non-required, user could still choose between `null` and `Action.ALLOW`.

azure-mgmt-appservice/src/test/resources/session-records/canCRUDFunctionApp.json
weidongxu-microsoft: kerb likely stands for "Kerberos keys", to include Kerberos keys in the response (when the storage account in AD identity-based authentication).

Personally never configured that...

weidongxu-microsoft: kerb likely stands for "Kerberos keys", to include Kerberos keys in the response (when the storage account in AD identity-based authentication).

Personally never configured that...

azure-mgmt-storage/src/main/java/com/azure/management/storage/models/BlobServicesInner.java
weidongxu-microsoft: Also since apiVersion and subscriptionId is filled internally, the method signature remain unchanged as the public client method. (method is service is not public hence user cannot use them)

PR #1151, opened 2020-03-26
Fix azure and azure-samples test for cosmos, aks, sql

ChenTanyi:


azure/src/test/java/com/azure/management/TestKubernetesCluster.java
```diff
@@ -27,8 +35,21 @@ public KubernetesCluster createResource(KubernetesClusters kubernetesClusters) t
...
+        // Thus, set it to AZURE_AUTH_LOCATION_2 when you want.
+        String envSecondaryServicePrincipal = System.getenv("AZURE_AUTH_LOCATION_2");
+        if (envSecondaryServicePrincipal == null || envSecondaryServicePrincipal.isEmpty() || !(new File(envSecondaryServicePrincipal).exists())) {
+            envSecondaryServicePrincipal = System.getenv("AZURE_AUTH_LOCATION");
+        }
+
+        try {
+            HashMap<String, String> credentialsMap = ParseAuthFile(envSecondaryServicePrincipal);
+            clientId = credentialsMap.get("clientId");
+            secret = credentialsMap.get("clientSecret");
```
weidongxu-microsoft: Please double check the recorded json for id and secret.

ChenTanyi: Yes, I've check for it. The clientId will be contained in record, and the secret will not. Secret is in the PUT body, and the record didn't record the request body.

ChenTanyi: Like previous in the AKS test.

Issue #1154, opened 2020-04-03
[FEATURE REQ] Update the getMasterKey() method of Azure Functions


Flanker32:
**Is your feature request related to a problem? Please describe.**
We want to use `FunctionApp.getMasterKey()` to get the key of function and trigger it, however, this call will failed with 409 error, which is a change of Function V2, here is the related [page](https://github.com/Azure/azure-functions-host/wiki/Changes-to-Key-Management-in-Functions-V2)

**Describe the solution you'd like**
We could use another API [Web Apps - List Host Keys](https://docs.microsoft.com/en-us/rest/api/appservice/webapps/listhostkeys) to get the function key, which could works for V2 functions.

**Describe alternatives you've considered**
Now we create an HTTP Post request to List Host Keys API by ourselves as a workaround.

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

PR #1157, opened 2020-04-08
update getMasterKey rest api

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1154

Backport from track2


weidongxu-microsoft:
@ChenTanyi

It is not in inner (at least not in this version). Either appservice forget to define it in swagger (happens a lot) or it is defined in later version.

The rest api is in use in portal.

Issue #1155, opened 2020-04-03
[FEATURE REQ] Provide triggerFunction method in SDK


Flanker32:
**Is your feature request related to a problem? Please describe.**
It could be better if we could have a trigger method in Function SDK, so that we could call this method to do the test job, especially for none Http triggers

**Describe the solution you'd like**
https://docs.microsoft.com/mt-mt/Azure/azure-functions/functions-manually-run-non-http Here is the official document for trigger non http triggers, we just need to post request to function admin api, it could more convenient if we have this api in SDK.

**Describe alternatives you've considered**
We are now post to the function admin api to trigger function by own code as a workaround.

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

PR #1158, opened 2020-04-08
support triggerFunction

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1155

Sample code
```
        Map<String, String> payload = new HashMap<>();
        payload.put("input", "testAAA");
        functionApp.triggerFunction("QueueTrigger1", payload);
```


weidongxu-microsoft:
@Flanker32 for review as well.

weidongxu-microsoft:
Tested, but it require function app + function deployed + download log to verify + not recordable, so no test added.

PR #1163, opened 2020-04-14
attempt migitation on bug in FlatteningSerializer

weidongxu-microsoft:
Attempt to fix https://github.com/Azure/azure-sdk-for-java/issues/10141

Only a mitigation.

Live test is good.


azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/implementation/TypeSerializationTests.java
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ *
+ * Code generated by Microsoft (R) AutoRest Code Generator.
```
ChenTanyi, line 6: I think it is not the generated code.

weidongxu-microsoft: Fixed.

PR #1177, opened 2020-05-06
Compute: support to set eviction policy in vmss

xccc-msft:
Fix: [Azure/azure-sdk-for-java#10706](https://github.com/Azure/azure-sdk-for-java/issues/10706)

### DIFF
--- a/azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachineScaleSet.java
+++ b/azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachineScaleSet.java
@@ -1724,6 +1724,14 @@ public interface VirtualMachineScaleSet extends
             WithCreate withVirtualMachinePriority(VirtualMachinePriorityTypes priority);
 
             /**
+             * Specifies the eviction policy of the virtual machines in the scale set.
+             *
+             * @param policy eviction policy for the virtual machines in the scale set.
+             * @return the next stage of the definition
+             */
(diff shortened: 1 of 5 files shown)


azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachineScaleSet.java
```diff
@@ -1739,8 +1739,25 @@ WithManagedCreate withNewDataDiskFromImage(int imageLun,
...
-            @Beta(Beta.SinceVersion.V1_11_0)
+            @Deprecated
             WithCreate withLowPriorityVirtualMachine(VirtualMachineEvictionPolicyTypes policy);
+
+            /**
+             * Specify that virtual machines in the scale set should be spot priority VMs.
+             *
+             * @return the next stage of the definition
+             */
+            WithCreate withSpotPriorityVirtualMachine();
```
weidongxu-microsoft, line 1750: Probably just `withSpotVirtualMachine`, as in doc?

xccc-msft: If so, should we set `withSpot()` instead of `withSpotPriority()` in VM?

weidongxu-microsoft: OK, `withSpot` seems not very intuitive as well. Then keep it as `withSpotPriorityVirtualMachine`.

Keep the PR open for a while to see if there is other opinions.

PR #1179, opened 2020-05-07
Resources: update defaultApiVersion with exact match only

xccc-msft:
Fix: [Azure/azure-sdk-for-java#10681](https://github.com/Azure/azure-sdk-for-java/issues/10681)
Align: [ResourceUtils::ResourceTypeForApiVersion](https://github.com/Azure/azure-libraries-for-net/blob/master/src/ResourceManagement/ResourceManager/Core/ResourceUtils.cs#L147-L157)


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/arm/ResourceUtils.java
```diff
@@ -152,27 +152,30 @@ public static String nameFromResourceId(String id) {
      * @return the default api version to use
      */
     public static String defaultApiVersion(String id, Provider provider) {
-        String resourceType = resourceTypeFromResourceId(id).toLowerCase();
+        String resourceType = resourceTypeForApiVersion(id);
```
wzur-r3, line 155: Azure has a habit to return resource ids with types in uppercase, so it would be beneficial to keep that
```suggestion
        String resourceType = resourceTypeForApiVersion(id).toLowerCase();
```

xccc-msft: It's using `equalsIgnoreCase` later in the comparing loop so I just remove this `toLowerCase`.

wzur-r3: my bad - I didn't look close enough

Issue #1181, opened 2020-05-14
[FEATURE REQ] AKS supports 2020-03-01 API version, especially new SKU "paid"


yungezz:
**Is your feature request related to a problem? Please describe.**
Customer is requesting AKS SDK supports 2020-03-01 API version, especially new SKU "paid". This should happen in Track1. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ ] Description Added
- [ ] Expected solution specified

PR #1182, opened 2020-05-15
Aks update: 2020 03

ChenTanyi:
close https://github.com/Azure/azure-libraries-for-java/issues/1181


azure-mgmt-containerservice/src/main/java/com/microsoft/azure/management/containerservice/KubernetesClusterAgentPool.java
```diff
@@ -222,6 +246,7 @@
             WithAgentPoolVirtualMachineCount<ParentT>,
             WithMaxPodsCount<ParentT>,
             WithVirtualNetwork<ParentT>,
+            WithMode<ParentT>,
```
weidongxu-microsoft, line 249: Is `agentPoolMode` mandatory? If so should it be in a separate stage?

azure-mgmt-containerservice/src/main/java/com/microsoft/azure/management/containerservice/KubernetesCluster.java
```diff
@@ -412,13 +416,27 @@
...
-             * Updates the cluster's add-on's profiles.
+             * Specifies the cluster's add-on's profiles.
              *
              * @param addOnProfileMap the cluster's add-on's profiles
-             * @return the next stage of the update
+             * @return the next stage of the definition
              */
             @Beta(SinceVersion.V1_15_0)
-            KubernetesCluster.Update withAddOnProfiles(Map<String, ManagedClusterAddonProfile> addOnProfileMap);
+            KubernetesCluster.DefinitionStages.WithCreate withAddOnProfiles(Map<String, ManagedClusterAddonProfile> addOnProfileMap);
```
xccc-msft, line 425: Does this change make `withAddOnProfiles` unavailable on update stage?

azure-mgmt-containerservice/src/main/java/com/microsoft/azure/management/containerservice/KubernetesClusterAgentPool.java
ChenTanyi: It is hard to say mandatory or not. The service requires one pool to be system and others could be default.

Do we prefer to use withSystemMode/withDefaultMode as a mandatory definition?

azure-mgmt-containerservice/src/main/java/com/microsoft/azure/management/containerservice/KubernetesCluster.java
ChenTanyi: There is another in update stage.

azure-mgmt-containerservice/src/main/java/com/microsoft/azure/management/containerservice/KubernetesClusterAgentPool.java
weidongxu-microsoft: Since the enum is `ExpandableStringEnum` so service could add new one anytime. In this case maybe just an enum would be good enough.

ChenTanyi: After discussing, I think adding it in the optional stage is enough. The service response is enough for customer.

Issue #1193, opened 2020-05-26
[FEATURE REQ] Support get tenant Id directly in Azure client


Flanker32:
**Is your feature request related to a problem? Please describe.**
We want to get the tenant id to auth with other mgmt library like application insights, however, it's difficult for us to get tenant id from Azure client.

**Describe the solution you'd like**
Azure provide a method like `getTenantId()` or expose more infos like `getAuthenticated()`

**Describe alternatives you've considered**
`azure.accessManagement().roleAssignments().manager().tenantId()`

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

PR #1194, opened 2020-05-26
feat: expose subs and tenants from resource manager and azure

ChenTanyi:
https://github.com/Azure/azure-libraries-for-java/issues/1193

### DIFF
--- a/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/ResourceManager.java
+++ b/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/ResourceManager.java
@@ -16,6 +16,7 @@ import com.microsoft.azure.management.resources.PolicyAssignments;
 import com.microsoft.azure.management.resources.PolicyDefinitions;
 import com.microsoft.azure.management.resources.Providers;
 import com.microsoft.azure.management.resources.ResourceGroups;
+import com.microsoft.azure.management.resources.Subscription;
 import com.microsoft.azure.management.resources.Subscriptions;
 import com.microsoft.azure.management.resources.Tenants;
 import com.microsoft.azure.management.resources.fluentcore.arm.AzureConfigurable;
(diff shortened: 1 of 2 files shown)


azure/src/main/java/com/microsoft/azure/management/Azure.java
```diff
@@ -437,6 +439,7 @@ private Azure(RestClient restClient, String subscriptionId, String tenantId, Aut
         this.monitorManager = MonitorManager.authenticate(restClient, subscriptionId);
         this.eventHubManager = EventHubManager.authenticate(restClient, subscriptionId);
         this.subscriptionId = subscriptionId;
+        this.tenantId = tenantId;
```
weidongxu-microsoft, line 442: If `authenticated` is available, use `authenticated.tenantId()` would save this variable?

ChenTanyi: Done

azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/ResourceManager.java
```diff
@@ -181,6 +186,26 @@ public ResourceGroups resourceGroups() {
         return resourceGroups;
     }
 
+    /**
+     * @return the subscription management API entry point
+     */
+    public Subscriptions subscriptions() {
```
xccc-msft, line 192: Do we need to expose `subscriptions()` and `tenants()` here? The existing logic is like we make them available inside `Authenticated` as it above the level of resource management.

ChenTanyi: Since it currently exposed in `Authenticated`, customer cannot use in the `Azure` object.

ChenTanyi: We could expose it by save the `Authenticated` object, but I think use a new Client is more clear.

PR #1220, opened 2020-06-15
Compute: support specialized gallery image version

xccc-msft:
Fix: [Azure/azure-sdk-for-java#11912](https://github.com/Azure/azure-sdk-for-java/issues/11912)

### DIFF
--- a/azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachine.java
+++ b/azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachine.java
@@ -809,7 +809,7 @@ public interface VirtualMachine extends
             WithWindowsAdminUsernameManaged withWindowsCustomImage(String customImageId);
 
             /**
-             * Specifies the resource ID of a generalized Windows gallery image version to be used as the virtual machine's OS.
+             * Specifies the resource ID of a Windows gallery image version to be used as the virtual machine's OS.
              *
              * @param galleryImageVersionId the resource ID of the gallery image version
              * @return the next stage of the definition
@@ -859,7 +859,7 @@ public interface VirtualMachine extends
             WithLinuxRootUsernameManaged withLinuxCustomImage(String customImageId);
 
             /**
-             * Specifies the resource ID of a generalized Linux gallery image version to be used as the virtual machines' OS.
+             * Specifies the resource ID of a Linux gallery image version to be used as the virtual machines' OS.
              *
              * @param galleryImageVersionId the resource ID of a gallery image version
              * @return the next stage of the definition
@@ -891,15 +891,6 @@ public interface VirtualMachine extends
              * @return the next stage of the definition
              */
             WithManagedCreate withSpecializedOSDisk(Disk disk, OperatingSystemTypes osType);
-
-            /**
-             * Specifies the resource ID of a specialized gallery image version to be used as the virtual machine's OS.
-             * The information of OS profile are not required.
-             *
-             * @param galleryImageVersionId the resource ID of the gallery image version
-             * @return the next stage of the definition
-             */
-            WithManagedCreate withSpecializedGalleryImageVersion(String galleryImageVersionId);
         }
 
         /**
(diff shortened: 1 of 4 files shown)


azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/implementation/VirtualMachineImpl.java
```diff
@@ -720,6 +722,16 @@ public VirtualMachineImpl withSpecializedOSDisk(Disk disk, OperatingSystemTypes
         return this;
     }
 
+    @Override
+    public VirtualMachineImpl withSpecializedGalleryImageVersion(String galleryImageVersionId) {
+        ImageReference imageReferenceInner = new ImageReference();
+        imageReferenceInner.withId(galleryImageVersionId);
+        this.inner().storageProfile().osDisk().withCreateOption(DiskCreateOptionTypes.FROM_IMAGE);
```
ChenTanyi, line 729: This makes `isOSDiskFromImage` true, so no need another bool.

xccc-msft: Only if it is `specialized`, we need to set OS profile as null. If it is `generalized`, keep as it is.

azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachine.java
```diff
@@ -913,6 +904,14 @@
...
              */
             WithLinuxRootPasswordOrPublicKeyManagedOrUnmanaged withRootUsername(String rootUserName);
+
+            /**
+             * Specifies to remove the information of OS profile for the Linux virtual machine.
+             * It is required when creating from a specialized image.
+             *
+             * @return the next stage of the definition
+             */
+            WithLinuxCreateManagedOrUnmanaged withoutRootUsernameAndPassword();
```
weidongxu-microsoft, line 914: Suggestion: Maybe same method like `withNoCredential` for both windows and linux flow?

xccc-msft: In vm/vmss, it should be OK as we already have existing methods like:
`withoutPrimaryPublicIPAddress`: the VM should not have a public IP address.
`withoutAutoUpdate`: disables automatic updates.

weidongxu-microsoft:
One question: why not have e.g.
```
WithWindowsCreateManagedOrUnmanaged withWindowsSpecializedGalleryImageVersion(String galleryImageVersionId);
```
which would skip the username/password stage?

xccc-msft:
> One question: why not have e.g.
> 
> ```
> WithWindowsCreateManagedOrUnmanaged withWindowsSpecializedGalleryImageVersion(String galleryImageVersionId);
> ```
> 
> which would skip the username/password stage?

It should also apply to other cases when passing specialized image(id or url)/image version. So might be better having `without*` as per my opinion.

weidongxu-microsoft:
> > One question: why not have e.g.
> > ```
> > WithWindowsCreateManagedOrUnmanaged withWindowsSpecializedGalleryImageVersion(String galleryImageVersionId);
> > ```
> > 
> > 
> > which would skip the username/password stage?
> 
> It should also apply to other cases when passing specialized image(id or url)/image version. So might be better having `without*` as per my opinion.

This appears to be the only method involving `Gallery`. Do we aware any other cases that having specialized image?

And we already have a `withSpecializedOSDisk` method. Please check whether this is for specialized OS.

xccc-msft:
> > > One question: why not have e.g.
> > > ```
> > > WithWindowsCreateManagedOrUnmanaged withWindowsSpecializedGalleryImageVersion(String galleryImageVersionId);
> > > ```
> > > 
> > > 
> > > which would skip the username/password stage?
> > 
> > 
> > It should also apply to other cases when passing specialized image(id or url)/image version. So might be better having `without*` as per my opinion.
> 
> This appears to be the only method involving `Gallery`. Do we aware any other cases that having specialized image?
> 
> And we already have a `withSpecializedOSDisk` method. Please check whether this is for specialized OS.

For methods like `withLinuxCustomImage(String customImageId)`, it may also have specialized. For `withSpecializedOSDisk`, it should be fine as the next phase doesn't require username/password input.

weidongxu-microsoft:
For `withLinuxCustomImage`, can you confirm that it supports specialized image? Any doc or tests?

For `withSpecializedOSDisk`, what I mean is that it appears already THE flow for specialized OS (as a existing OS disk -- apparently specialized, instead of a specialized image).

xccc-msft:
> For `withLinuxCustomImage`, can you confirm that it supports specialized image? Any doc or tests?
> 
> For `withSpecializedOSDisk`, what I mean is that it appears already THE flow for specialized OS (as a existing OS disk -- apparently specialized, instead of a specialized image).

The specialized/generialized is set on image level. Under the image, you could have different versions like `0.0.1`/`0.0.2`, which is the image version reported by our user.
The OS profile would only be passed when importing image. When importing OS disk, the OS profile will be set as null.

PR #1224, opened 2020-06-16
Upgrade guava and client-runtime

weidongxu-microsoft:


weidongxu-microsoft:
revert guava upgrade, 20.0 is the last version that supports java 7...

PR #1237, opened 2020-07-15
Resources: upgrade to package-2020-06

xccc-msft:
1. Support tag operations
2. Support outResources in deployment


azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/implementation/TypeSerializationTests.java
```diff
@@ -29,7 +29,7 @@ private static DeploymentInner createRequestFromInner(DeploymentImpl deployment)
...
..."VNet name\"}},\"vnetAddressPrefix\":{\"type\":\"string\",\"defaultValue\":\"10.0.0.0/16\",\"metadata\":{\"description\":\"Address prefix\"}},\"subnet1Prefix\":{\"type\":\"string\",\"defaultValue\":\"10.0.0.0/24\",\"metadata\":{\"description\":\"Subnet 1 Prefix\"}},\"subnet1Name\":{\"type\":\"string\",\"defaultValue\":\"Subnet1\",\"metadata\":{\"description\":\"Subnet 1 Name\"}},\"subnet2Prefix\":{\"type\":\"string\",\"defaultValue\":\"10.0.1.0/24\",\"metadata\":{\"description\":\"Subnet 2 Prefix\"}},\"subnet2Name\":{\"type\":\"string\",\"defaultValue\":\"Subnet222\",\"metadata\":{\"description\":\"Subnet 2 Name\"}}},\"variables\":{\"apiVersion\":\"2015-06-15\"},\"resources\":[{\"apiVersion\":\"[variables('apiVersion')]\",\"type\":\"Microsoft.Network/virtualNetworks\",\"name\":\"[parameters('vnetName')]\",\"location\":\"[resourceGroup().location]\",\"properties\":{\"addressSpace\":{\"addressPrefixes\":[\"[parameters('vnetAddressPrefix')]\"]},\"subnets\":[{\"name\":\"[parameters('subnet1Name')]\",\"properties\":{\"addressPrefix\":\"[parameters('subnet1Prefix')]\"}},{\"name\":\"[parameters('subnet2Name')]\",\"properties\":{\"addressPrefix\":\"[parameters('subnet2Prefix')]\"}}]}}]}");
```
xccc-msft, line 32: This is not valid change as template got removed from return type. Need to check with @weidongxu-microsoft  if we can have alternative inner classes.

azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/GenericResourcesImpl.java
```diff
@@ -35,29 +37,37 @@
...
-    GenericResourcesImpl(ResourceManager resourceManager) {
+    private PagedListConverter<GenericResourceExpandedInner, GenericResource> converter;
+
+    GenericResourcesImpl(final ResourceManager resourceManager) {
         super(resourceManager.inner().resources(), resourceManager);
+        converter = new PagedListConverter<GenericResourceExpandedInner, GenericResource>() {
+            @Override
+            public Observable<GenericResource> typeConvertAsync(GenericResourceExpandedInner inner) {
+                return Observable.just((GenericResource) new GenericResourceImpl(inner.name(), inner, manager()));
+            }
```
weidongxu-microsoft, line 48: No. This losses information.

If service do this intentionally, the right approach is to had `GenericResource` wraps `GenericResourceExpandedInner` (when by get, there additional properties be null).

weidongxu-microsoft: It might be OK if we determined not to maintain it anymore...

weidongxu-microsoft: Maybe it is OK.

Even if we later need to expose createdTime etc., we can hack by first check inner instanceof GenericResourceExpandedInner , if yet, cast and return value, if not, return null.

xccc-msft: Now I'm not sure if we should expose filter/expand in APIs. For the solution, I agree we can do in that way by checking instanceof.

weidongxu-microsoft: Generally speaking, we should expose as overloaded method. But for track1, I think we will not expose new ones, until requested by customer.

For track2, I think the question is whether we expose a simple string (and let user compose it based on description), or try standardize it a bit.

We likely need more data on existing ones.
E.g. description on compute
```
     * @param filter The filter to apply to the operation. Allowed values are 'startswith(instanceView/statuses/code,
     *     'PowerState') eq true', 'properties/latestModelApplied eq true', 'properties/latestModelApplied eq false'.
```
So I think most is using odata $filter $expand. We need first investigate on existing support on odata.

azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/implementation/TypeSerializationTests.java
xccc-msft: Marked by Ignore to skip for now.

azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/DeploymentImpl.java
```diff
@@ -355,16 +354,6 @@ public DeploymentImpl beginCreate() {
 
     @Override
     public Observable<Deployment> updateResourceAsync() {
```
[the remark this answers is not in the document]
xccc-msft: Deleted.

azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/implementation/DeploymentImpl.java
```diff
@@ -58,13 +58,15 @@
     private Creatable<ResourceGroup> creatableResourceGroup;
     private ObjectMapper objectMapper;
     private DeploymentWhatIf deploymentWhatIf;
+    private DeploymentInner deploymentInner;
```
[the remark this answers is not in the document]
xccc-msft: Done.

PR #1238, opened 2020-07-20
NP protection

weidongxu-microsoft:
fix https://github.com/Azure/azure-sdk-for-java/issues/13316

### DIFF
--- a/azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/implementation/MetricDefinitionsImpl.java
+++ b/azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/implementation/MetricDefinitionsImpl.java
@@ -41,10 +41,14 @@ class MetricDefinitionsImpl
 
     @Override
     public List<MetricDefinition> listByResource(String resourceId) {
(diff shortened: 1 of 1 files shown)


azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/implementation/MetricDefinitionsImpl.java
```diff
@@ -41,10 +41,14 @@ public MetricDefinitionsInner inner() {
 
     @Override
     public List<MetricDefinition> listByResource(String resourceId) {
-        return Lists.transform(this.inner().list(resourceId), new Function<MetricDefinitionInner, MetricDefinition>() {
+        List<MetricDefinitionInner> inners = this.inner().list(resourceId);
+        if (inners == null) {
+            return null;
```
ArchanaAnand0212, line 46: Is there a reason we are not returning empty list instead of null? 
End client still have to put null check in place for this.

weidongxu-microsoft: Note that this lib is deprecated.

For legacy code, it usually be such. In more modern code, it usually returns a `Collections.emptyList()`.

PR #1240, opened 2020-07-22
add auxiliary header directly for cross tenant

ChenTanyi:


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/AuxiliaryCredentialsInterceptor.java
```diff
@@ -42,36 +42,21 @@ public AuxiliaryCredentialsInterceptor(AzureTokenCredentials... credentials) {
     public Response intercept(Chain chain) throws IOException {
         Response response = chain.proceed(chain.request());
         if (!response.isSuccessful() && tokenCredentials != null && this.tokenCredentials.length > 0) {
-            String content = errorBody(response.body());
-            AzureJacksonAdapter jacksonAdapter = new AzureJacksonAdapter();
-            CloudError cloudError = jacksonAdapter.deserialize(content, CloudError.class);
-            if (cloudError != null && LINKED_AUTHORIZATION_FAILED.equals(cloudError.code())) {
```
xccc-msft, line 48: For DeploymentFailed, the `LINKED_AUTHORIZATION_FAILED` could be found via `cloudError.details().get(index).code()`. It would be better to traverse inside `details()` if the `code()` is not `LINKED_AUTHORIZATION_FAILED`, instead of deleting the check logic directly.

ChenTanyi: `LINKED_AUTHORIZATION_FAILED` needs to parse `details().get(index).message()`, rather than in `code()`

weidongxu-microsoft: The problem maybe how much we can trust service error message and its detail, and whether service might change it in future, without our knowledge. Hence it probably much safer to just go with the auxiliary if it is provided.

PR #1249, opened 2020-07-29
update vm images

weidongxu-microsoft:
Fix https://msazure.visualstudio.com/Unified%20Platform%20KPIs/_workitems/edit/6476653


weidongxu-microsoft:
> Better to add a live only test for the image.

It is track1, so I only do minimal. I would now count on track2 live tests to find out-dated images ;-)

Issue #1251, opened 2020-07-31
[FEATURE REQ] investigate potential improvement on ResourceManagerThrottlingInterceptor


weidongxu-microsoft:
E.g. throttling on PUT should not block DELETE

PR #1252, opened 2020-07-31
throttle base on subscriptionId and request method

weidongxu-microsoft:
fix https://github.com/Azure/azure-libraries-for-java/issues/1251

ratelimit could have different limit on write/read/delete
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling

Hence it likely helps to separate them (to avoid the case that a throttle write blocks a subsequent delete or read).

Pending mock server test.

### DIFF
--- a/azure-mgmt-resources/pom.xml
+++ b/azure-mgmt-resources/pom.xml
@@ -84,6 +84,12 @@
       <artifactId>httpcore</artifactId>
       <version>4.4.5</version>
     </dependency>
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock-standalone</artifactId>
+      <version>2.24.1</version>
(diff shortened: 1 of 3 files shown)


weidongxu-microsoft:
The PR should work for compute RP as well. They also have different scope for PUT/DELETE

```
[main] INFO com.microsoft.azure.management.compute.Disks createOrUpdate - x-ms-ratelimit-remaining-resource: Microsoft.Compute/CreateUpdateDisks3Min;999,Microsoft.Compute/CreateUpdateDisks30Min;7999

[main] INFO com.microsoft.azure.management.compute.Disks delete - x-ms-ratelimit-remaining-resource: Microsoft.Compute/DeleteDisks3Min;999,Microsoft.Compute/DeleteDisks30Min;7999
```

PR #1254, opened 2020-08-05
ReturnRequestIdHeaderInterceptor

weidongxu-microsoft:
to request server put x-ms-client-request-id in response

Sample usage:

```
        Azure azure = Azure.configure()
                .withInterceptor(new ReturnRequestIdHeaderInterceptor())
                .authenticate(...)
                ...
```

Header can be accessed in Exception.

```
        exception.response().headers().get("x-ms-client-request-id")
```

### DIFF
--- a/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ReturnRequestIdHeaderInterceptor.java
+++ b/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ReturnRequestIdHeaderInterceptor.java
@@ -1,0 +1,69 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure.management.resources.fluentcore.utils;
+
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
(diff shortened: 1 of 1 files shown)


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ReturnRequestIdHeaderInterceptor.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String NAME_CLIENT_REQUEST_ID = "x-ms-client-request-id";
+
+    private final boolean copyClientRequestIdInResponse;
+
+    /**
+     * Creates a new instance of ReturnRequestIdHeaderInterceptor.
+     * Sets "x-ms-return-client-request-id: true" in requests headers.
+     */
+    public ReturnRequestIdHeaderInterceptor() {
+        this(false);
```
ChenTanyi, line 31: Might better default `true` or just remove this boolean.

weidongxu-microsoft: One reason customer need the "x-ms-client-request-id" is to cross-check with service when something went wrong. If we send "x-ms-return-client-request-id=true" but service does not return request id, it might be an indicate that service is not handling "x-ms-client-request-id".

PR #1264, opened 2020-08-20
fix function app slot

weidongxu-microsoft:
@Flanker32 

3 issues.

1. service not accept the empty siteConfig in PUT when create new slot
2. PATCH not working, had to use PUT
3. SDK error in sending name in request (name from service is e.g. `java-funcapp-36626c/java-slot-43411876b`, which is `parent/slot`).

### DIFF
--- a/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/FunctionDeploymentSlotImpl.java
+++ b/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/FunctionDeploymentSlotImpl.java
@@ -10,6 +10,7 @@ import com.microsoft.azure.management.apigeneration.LangDefinition;
 import com.microsoft.azure.management.appservice.FunctionApp;
 import com.microsoft.azure.management.appservice.FunctionDeploymentSlot;
 import com.microsoft.azure.management.appservice.FunctionDeploymentSlot.DefinitionStages.WithCreate;
+import com.microsoft.azure.management.appservice.SitePatchResource;
 import rx.Completable;
 import rx.Observable;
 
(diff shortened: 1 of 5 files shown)


azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/FunctionDeploymentSlotImpl.java
```diff
@@ -72,4 +74,15 @@ public Completable zipDeployAsync(File zipFile) {
             return Completable.error(e);
         }
     }
-}
\ No newline at end of file
+
+    @Override
+    Observable<SiteInner> submitSite(final SiteInner site) {
+        return submitSiteWithoutSiteConfig(site);
```
ChenTanyi, line 80: the webapp base has `site.withSiteConfig(new SiteConfig());`. Does it need to be added here?

ChenTanyi (approved):

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/FunctionDeploymentSlotImpl.java
weidongxu-microsoft: Service will return this, if set...

```
Function App app settings are not expected to be empty. Please try again with WEBSITE_CONTENTAZUREFILECONNECTIONSTRING as a storage account connection string.: Function App app settings are not expected to be empty. Please try again with WEBSITE_CONTENTAZUREFILECONNECTIONSTRING as a storage account connection string.
```

weidongxu-microsoft:
`withBrandNewConfiguration` will not work for function app slot.

In track2 I might just delete this choice.

PR #1272, opened 2020-09-04
use reserved to decide OS. "kind" of asp no longer has "linux" in it

weidongxu-microsoft:


ChenTanyi (approved):

xccc-msft (approved):

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/AppServicePlans.java
```diff
@@ -35,4 +37,20 @@
...
         HasManager<AppServiceManager>,
         HasInner<AppServicePlansInner> {
+
+    /**
+     * Lists all the resources of the specified type in the currently selected subscription.
+     *
+     * @param includeDetails include standard details
+     * @return list of resources
+     */
+    PagedList<AppServicePlan> list(boolean includeDetails);
```
weidongxu-microsoft, line 47: @Flanker32 you need to call this `list` with `true`.

PR #1289, opened 2020-10-09
Compute: support virtual machines force delete

xccc-msft:
1. Upgrade api version to package-2020-06-30, which supports force delete
2. Run live tests to update session records after upgrade
3. Add new interfaces for force delete
4. Add new test cases for force delete


azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/VirtualMachines.java
```diff
@@ -341,4 +341,55 @@
      * @return handle to the asynchronous execution
      */
     Observable<RunCommandResult> runCommandAsync(String groupName, String name, RunCommandInput inputCommand);
+
+    /**
+     * Force delete a resource from Azure, identifying it by its resource ID.
+     *
+     * @param id the resource ID of the resource to delete
+     */
+    void forceDeleteById(String id);
```
weidongxu-microsoft, line 350: Personally I prefer overload of `deleteById(id, force)` a little bit.

You can keep this unless anyone else got preference as well.

weidongxu-microsoft (approved):

Issue #1306, opened 2020-11-02
[BUG] Auto created storage account name may be invalid in FunctionAppImpl


Flanker32:
**Describe the bug**
`FunctionAppImpl` will auto create storage account with function name as prefix([refers](https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/FunctionAppImpl.java#L653)), however, function name may contains "-" and this may be include in the `randomName` result([refers](https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/fluentcore/utils/ResourceNamer.java#L38-L58)) which could not be used in storage account name

***Exception or Stack Trace***
```
Exception in thread "main" com.microsoft.azure.CloudException: test-appef78792216 is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only.: test-appef78792216 is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only.
	at com.microsoft.azure.AzureClient.createExceptionFromResponse(AzureClient.java:771)
	at com.microsoft.azure.AzureClient.access$100(AzureClient.java:33)
	at com.microsoft.azure.AzureClient$3.call(AzureClient.java:160)
	at com.microsoft.azure.AzureClient$3.call(AzureClient.java:157)
	at rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:69)
	at retrofit2.adapter.rxjava.CallArbiter.deliverResponse(CallArbiter.java:120)
	at retrofit2.adapter.rxjava.CallArbiter.emitResponse(CallArbiter.java:102)
	at retrofit2.adapter.rxjava.CallExecuteOnSubscribe.call(CallExecuteOnSubscribe.java:46)
	at retrofit2.adapter.rxjava.CallExecuteOnSubscribe.call(CallExecuteOnSubscribe.java:24)
	at rx.Observable.unsafeSubscribe(Observable.java:10327)
	at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:48)
	at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:33)
	at rx.Observable.unsafeSubscribe(Observable.java:10327)
[...]

PR #1308, opened 2020-11-03
remove not allowed char in storage account name

weidongxu-microsoft:
fix https://github.com/Azure/azure-libraries-for-java/issues/1306


ChenTanyi (approved):

azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/implementation/FunctionAppImpl.java
```diff
@@ -664,6 +664,10 @@ public Completable afterPostRunAsync(final boolean isGroupFaulted) {
         return super.afterPostRunAsync(isGroupFaulted);
     }
 
+    private String getStorageAccountName() {
+        return name().replaceAll("[^a-zA-Z0-9]", "");
```
weidongxu-microsoft, line 668: uppercase is fine here, it will be turned to lower case in `randomResourceName`

xccc-msft (approved):

Issue #1314, opened 2020-11-16
[FEATURE REQ] support get AppSetting from Kudu


weidongxu-microsoft:
endpoint “/api/settings"

sample response
```
{"deployment_branch":"master","SCM_TRACE_LEVEL":"1","SCM_COMMAND_IDLE_TIMEOUT":"60","SCM_LOGSTREAM_TIMEOUT":"7200","SCM_BUILD_ARGS":"","aspnet:PortableCompilationOutput":"true","aspnet:PortableCompilationOutputSnapshotType":"Microsoft.Web.Compilation.Snapshots.SnapshotHelper, Microsoft.Web.Compilation.Snapshots, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35","aspnet:DisableFcnDaclRead":"true","SCM_GIT_USERNAME":"windowsazure","SCM_GIT_EMAIL":"windowsazure","webpages:Version":"3.0.0.0","webpages:Enabled":"true","webactivator:assembliesToScan":"Kudu.Services.Web","WEBSITE_SLOT_NAME":"Production","FUNCTIONS_EXTENSION_VERSION":"~3","ScmType":"None","WEBSITE_AUTH_ENABLED":"False","WEBSITE_CONTENTAZUREFILECONNECTIONSTRING":"###","WEBSITE_CONTENTSHARE":"fa1weidxu8735","FUNCTIONS_WORKER_RUNTIME":"java","WEBSITE_SITE_NAME":"fa1weidxu","AzureWebJobsStorage":"###"}
```

PR #1315, opened 2020-11-16
webapp support api/settings from Kudu

weidongxu-microsoft:
fix https://github.com/Azure/azure-libraries-for-java/issues/1314


weidongxu-microsoft:
@Flanker32 for review.

I didn't get a good name for the API method. Currently call it `getSiteAppSettings` (or maybe `getKuduAppSettings`). Let me know if any better idea.

Flanker32:
> @Flanker32 for review.
> 
> I didn't get a good name for the API method. Currently call it `getSiteAppSettings` (or maybe `getKuduAppSettings`). Let me know if any better idea.

@weidongxu-microsoft Thanks a lot for your efforts! I think `getKuduAppSettings` may be better but both are fine~

xccc-msft (approved):

weidongxu-microsoft:
We will still use `getSiteAppSettings`, since previously we did not expose any method with "Kudu" in name (logStream, or zipDeploy).

Issue #1348, opened 2021-01-22
[FEATURE REQ] Avoid sending redundant PATCH request to VM when creating extension - Databricks


weidongxu-microsoft:
Create VM extension on VM update flow without PATCH to VM itself (if no modify to VM)

PR #1350, opened 2021-01-25
Avoid PATCH during VM update flow, when VM itself is not modified

weidongxu-microsoft:
fix https://github.com/Azure/azure-libraries-for-java/issues/1348

### DIFF
--- a/azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/core/TestUtilities.java
+++ b/azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/core/TestUtilities.java
@@ -6,6 +6,18 @@
 
 package com.microsoft.azure.management.resources.core;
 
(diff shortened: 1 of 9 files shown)


azure-mgmt-resources/src/test/java/com/microsoft/azure/management/resources/core/TestUtilities.java
```diff
@@ -27,4 +39,32 @@ public static void sleep(int milliseconds, boolean isRecordMode) {
     public static String createPassword() {
         return "12NewPA$$w0rd!";
     }
+
+    private static String sshPublicKey;
+
+    public static String createSshPublicKey() {
```
weidongxu-microsoft, line 45: @xccc-msft I've added a method to generate public ssh key.

The public key will appear in response json. Hope CredScan is ok with it.

xccc-msft (approved):

PR #1351, opened 2021-01-26
regen subscriptions, add properties for Location

weidongxu-microsoft:


azure-mgmt-resources/src/main/java/com/microsoft/azure/management/resources/Location.java
```diff
@@ -45,4 +45,24 @@
...
+    /**
+     * @return the geography group.
+     */
+    String geographyGroup();
+
+    /**
+     * @return the physical location.
+     */
+    String physicalLocation();
 }
```
weidongxu-microsoft, line 68: Properties added.

shenqianjin: 👍

xccc-msft (approved):

PR #1354, opened 2021-01-29
force deletion on resource group

weidongxu-microsoft:
https://github.com/Azure/azure-libraries-for-java/pull/1354/commits/a91bf8a83c6cebb815455f890968e08ab42c5ee2 is the code change.


xccc-msft (approved):
Have this been tested on local?

weidongxu-microsoft:
> Have this been tested on local?

Yes, but no way to tell whether it is "force" or now, just see the query parameter is on url.

xccc-msft:
> > Have this been tested on local?
> 
> Yes, but no way to tell whether it is "force" or now, just see the query parameter is on url.

Not sure if it gets supported all regions now. For compute force delete, it could throw errors when testing on unsupported regions.

weidongxu-microsoft:
> > > Have this been tested on local?
> > 
> > 
> > Yes, but no way to tell whether it is "force" or now, just see the query parameter is on url.
> 
> Not sure if it gets supported all regions now. For compute force delete, it could throw errors when testing on unsupported regions.

Yes, I saw that on VM delete.

But apparently this cannot be for resource group delete. This need to be successful, regardless force or not, EUAP or not. (delete is success on normal region)

PR #1356, opened 2021-02-02
Batch: upgrade to API version 2021-01-01

xccc-msft:
1. Upgrade to 2021-01-01
2. Add `Deprecate` annotation for `maxTasksPerNode`. Add new `taskSlotsPerNode`.
3. Run live tests and update session records.


azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/implementation/ApplicationPackagesInner.java
```diff
@@ -276,7 +276,7 @@ public ApplicationPackageInner call(ServiceResponse<ApplicationPackageInner> res
         if (this.client.apiVersion() == null) {
             throw new IllegalArgumentException("Parameter this.client.apiVersion() is required and cannot be null.");
         }
-        return service.create(resourceGroupName, accountName, applicationName, versionName, this.client.subscriptionId(), this.client.apiVersion(), this.client.acceptLanguage(), parameters, this.client.userAgent())
+        return service.create(resourceGroupName, accountName, applicationName, versionName, this.client.subscriptionId(), this.client.apiVersion(), this.client.acceptLanguage(), null, this.client.userAgent())
```
weidongxu-microsoft: Why it is null?

xccc-msft: Fixed in later commit by using default constructor.

weidongxu-microsoft: But this should be the parameter that need to be passed in, isn't it?

xccc-msft: No, according to above links, only name is required rather than payload. And double checked how we define a new package [here](https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-batch/src/main/java/com/microsoft/azure/management/batch/Application.java#L81)

weidongxu-microsoft: OK, I see. All its properties is readonly. Maybe this is the reason this object is not in parameter.

weidongxu-microsoft (approved):

gingi (approved):

gingi:
Wanted to confirm that `taskSlotsPerNode` is a renaming of `maxTasksPerNode`.

PR #1385, opened 2021-03-30
fix update flow in graph-rbac

weidongxu-microsoft:
Fix https://github.com/Azure/azure-sdk-for-java/issues/20161


ChenTanyi (approved):

ChenTanyi (commented):
I think this should be better called in `update()`

weidongxu-microsoft:
Both of the "without" method only available in "update()" flow. So it should be safe.

ChenTanyi (approved):

PR #1401, opened 2021-07-08
Update pom.xml and put a more obvious message

nickzhums:


weidongxu-microsoft (approved):

ChenTanyi:
Might be better to change all pom, like https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-resources/pom.xml#L19

nickzhums:
could you mege the PR ? I don't  have permission @weidongxu-microsoft  thanks!

nickzhums:
> Might be better to change all pom, like https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-resources/pom.xml#L19

hmm good point, is there a quick way to do this?

ChenTanyi:
@nickzhums You can do text replacement, I think all messages are same among all pom.

